PR #1, opened 2018-11-07
Initial URL datasource based on http4s

wemrysi:
[ch2705]


core/src/test/scala/quasar/datasource/url/Http4sUrlDatasourceModuleSpec.scala
```diff
@@ -0,0 +1,232 @@
...
+
+      dsmod.lightweightDatasource[IO](config).map(_.toEither must beLike {
+        case Left(DatasourceError.ConnectionFailed(_, c, err)) =>
+          c must_=== config
+          err.getMessage must contain("Not Found")
+      }).unsafeRunSync()
+    }
+  }
+
+  "operation" >> {
```
alissapajer: I found myself thinking about potentially malicious entry points. In this case, we're downloading whatever the url points to (even if it's not json) and attempting to parse it. We'll fail to parse if it's not json, but we still downloaded whatever it was. Maybe it's just up to the user not to point to anything harmful and not on us if they do?

core/src/test/scala/quasar/datasource/url/MockClientBuilder.scala
```diff
@@ -0,0 +1,46 @@
...
+
+import scala.concurrent.ExecutionContext
+
+import cats.syntax.functor._
+import cats.effect.{ConcurrentEffect, Resource}
+import cats.effect.concurrent.Ref
+import org.http4s.{Response, Status}
+import org.http4s.client.Client
+
+object MockClientBuilder {
```
alissapajer: This looks like it would be useful as part of a testkit published by `http4s`.

core/src/main/scala/quasar/datasource/url/ResourceType.scala
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package quasar.datasource.url
+
+import slamdata.Predef.{Product, Serializable, String}
+
+import argonaut._, Argonaut._
+import scalaz.{Equal, Show}
+
+sealed trait ResourceType extends Product with Serializable
```
alissapajer: Should `ResourceType` be something that `quasar` provides so that the types are standardized across all datasource implementations? Maybe there's not a use for it outside of this datasource though.

core/src/test/scala/quasar/datasource/url/Http4sUrlDatasourceModuleSpec.scala
djspiewak: It's definitely on us. This basically leads to two questions: what sort of harm can be caused by downloading and attempting to parse, and how can we mitigate that harm? I don't think there are any viable executable exploits, since we're very memory-safe. We could be DoSed or crashed though, given our current handling of certain variants of wide data. We should make sure that doesn't happen, though, all things considered I suppose that could be a bit on the user since they effectively asked us to do something weird.

I think there's nothing that can be done here that is *worse* than what would happen if the user put the same data into S3.

They also can't, so far as I can tell, use this as a mechanism to launch DoS attacks in any way. We're just… running a single request, and not even in a crazy retry loop. So I think we're safe on that front.

core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
```diff
@@ -0,0 +1,188 @@
...
+
+  def sanitizeConfig(config: Json): Json = {
+    val sanitized = for {
+      headers <- config.cursor --\ "headers"
+
+      headerNames <- headers.fields
+
+      sensitiveNames = headerNames.filter(n => SensitiveHeaders.contains(n.ci))
+
+      redacted = sensitiveNames.foldLeft(headers) { (c, n) =>
```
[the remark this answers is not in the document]
wemrysi: It would, if this was a `Headers` instance, unfortunately it's a `Json` object.

core/src/test/scala/quasar/datasource/url/MockClientBuilder.scala
wemrysi: Aye, so would the `Http4sClientBuilder`! They have one for servers, but not clients.

core/src/main/scala/quasar/datasource/url/ResourceType.scala
wemrysi: Are you referring to the semantics or the string encoding? This mainly exists to hang a json encoding on, `ParsableType` is the standardized semantic type (which we translate this into).

wemrysi: Maybe we can add an `EncodeJson[ParsableType]` over in quasar and use that here and elsewhere?

alissapajer: Adding `EncodeJson[ParsableType]` would achieve what I was suggesting. That way we're not translating between two types which are encoding the same information.

alissapajer: If we want to do that, it can be done separately (since it'll require a new quasar version anyway).

core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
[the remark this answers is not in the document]
alissapajer: :( oh well

core/src/test/scala/quasar/datasource/url/Http4sUrlDatasourceModuleSpec.scala
alissapajer: Limiting max redirects, like we do, also gives the user more control since they can set it to `0` if they know they're pointing somewhere that doesn't need any redirects. Actually @wemrysi do you think that would be a better default than `5`?

wemrysi: @alissapajer hm, it seems theoretically more secure, but at the cost of convenience given redirects are pretty common. i.e. you'd create your datasource, fail due to redirects, have to reconfigure, etc.

alissapajer: Fair enough. If someone is concerned about redirects then they can set it to `0`.

PR #53, opened 2018-12-15
interpreted-read update

cryogenian:
depends on slamdata/quasar#3989


core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
```diff
@@ -120,12 +122,19 @@ class Http4sUrlDatasourceModule(clientBuilder: Http4sClientBuilder)
...
 
           urlConfig.compressionScheme.fold(qr)(QueryResult.compressed(_, qr))
         }
-
-        SingletonDatasource[F, Stream[F, ?], QueryResult[F]](kind, location, queryResult)
+        val ds = SingletonDatasource[F, Stream[F, ?], QueryResult[F]](kind, location, queryResult)
+        val preserveInstructions: (DS[F] => DS[F]) =
+          Datasource.evaluator[F, Stream[F, ?], InterpretedRead[ResourcePath], QueryResult[F]] modify { qe =>
+            QueryEvaluator(iRead => qe.evaluate(iRead).map(_ match {
+              case QueryResult.Typed(a, b, _) => QueryResult.Typed(a, b, iRead.instructions)
```
alissapajer: Why preserve the original instructions here? Shouldn't a datasource do the right thing by returning a `QueryResult` that already contains the uninterpreted instructions?

cryogenian: Right, but now the datasource does nothing with instructions and so, remaining instructions === original instructions.

cryogenian: Or you mean that `datasource` should return input parse instructions? I didn't find the way to get it here other than lensing because there is no `evaluate` in this module.

wemrysi: So I think the impedance mismatch here is that `SingletonDatasource` shouldn't have `InterpretedRead[ResourcePath]` as its query type, it should stick with just `ResourcePath`. I didn't catch this in slamdata/quasar#3989.

If we revert that, then this change via the lens would make more sense, I think.

alissapajer: Why do we have to make this lensed change at all though? If the datasource isn't using the instructions, wouldn't the `QueryResult`'s `List[ParseInstruction]` already be the original list of instructions?

cryogenian: So, what do you suggest to do? Go back to quasar and change `SingletonDatasource` signature? Maybe remove it out of quasar and introduce here?

cryogenian: @alissapajer Because query result is constructed here and `SingletonDatasource` simply returns it w/o modification and here we have no instructions. This is other possibility, I can make `SingletonDatasource` returning only `QueryResult` not any `R`

core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
```diff
@@ -52,6 +54,9 @@ import shims._
 class Http4sUrlDatasourceModule(clientBuilder: Http4sClientBuilder)
     extends LightweightDatasourceModule {
 
+  case class InternalResult[F[_]](data: Stream[F, Byte], compressed: Option[CompressionScheme], parseType: JsonVariant)
```
wemrysi: Should be `private` also.

PR #130, opened 2019-03-27
Support multiple urls

rintcius:
[ch6217]


core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
```diff
@@ -195,12 +209,21 @@ class Http4sUrlDatasourceModule(clientBuilder: Http4sClientBuilder)
       .filterNot(_.isEmpty)
       .value
 
-  private def validateUrl[F[_]: Sync](
+  private def validateRequests[F[_]: Sync](
+      client: Client[F],
+      requests: IList[UriRequest])
+      : F[Condition[InitializationError[Unit]]] =
+    requests.foldLeft(Condition.normal[InitializationError[Unit]]().point[F]) { case (acc, req) =>
```
rintcius: may be worth to do this in parallel

core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
```diff
@@ -116,20 +99,19 @@ class Http4sUrlDatasourceModule(clientBuilder: Http4sClientBuilder)
...
-            case ResourceType.Json => JsonVariant.ArrayWrapped
-            case ResourceType.LdJson => JsonVariant.LineDelimited
-          }
-          InternalResult(body.onFinalize(d.dispose), urlConfig.compressionScheme, jvar)
-        }
+        val content: IMap[ResourcePath, F[InternalResult[F]]] =
+          urlConfig.requests.zipWithIndex.foldLeft(IMap.empty[ResourcePath, F[InternalResult[F]]])
+            { case (acc, (req, i)) =>
+              val n: ResourceName =
+                name(req.url).getOrElse(defaultName(urlConfig.resourceType, i))
```
rintcius: The resource paths that are generated is the path of the url, and if that's empty the hostname. If both are empty then a default name of the form `url_${i}.(ld)json` is used. If there are duplicates then the names will be prefixed with a `_`. So this will always result in resources directly under root.

We can consider other templates though. E.g. 2 level resource paths of the form `hostname:port / urlPath` (and take care of non-existing hosts, ports and paths). This will group by `hostname:port` which may be useful in some scenarios.

Adding more levels could work too, but there's a trade-off because adding levels also results in more clicks in the UI.

rintcius:
Depends on https://github.com/slamdata/quasar/pull/4116

core/src/main/scala/quasar/datasource/url/UrlDatasourceConfig.scala
```diff
@@ -16,35 +16,55 @@
...
 import scalaz.std.option._
 import scalaz.syntax.traverse._
 import shims._
 
-final case class UrlDatasourceConfig(
+final case class UriRequest(
     url: Uri,
+    headers: IMap[CaseInsensitiveString, String])
+
+final case class UrlDatasourceConfig(
```
wemrysi: Hm, I wonder if we should have `resourceType` and `compressionScheme` be per-URL as well.

core/src/main/scala/quasar/datasource/url/UrlDatasourceConfig.scala
```diff
@@ -16,35 +16,55 @@
...
     resourceType: ResourceType,
     compressionScheme: Option[CompressionScheme],
-    headers: IMap[CaseInsensitiveString, String],
     maxRedirects: Int)
 
 object UrlDatasourceConfig {
   val DefaultMaxRedirects: Int = 5
 
-  implicit val decodeJson: DecodeJson[UrlDatasourceConfig] =
+  implicit val decodeUrl: DecodeJson[Uri] = DecodeJson.StringDecodeJson.flatMap(s =>
```
wemrysi: `decodeUrl` and `decodeHeaders` should probably be `private`.

core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
wemrysi: How about just always making it `hostname/path` in order to avoid path conflicts?

core/src/main/scala/quasar/datasource/url/UrlDatasourceConfig.scala
rintcius: Yeah, I didn't like the duplication that would involve in entering the config, My thought was that a customer would typically use 1 combination, and if they don't then it wouldn't be too bad to create a new datasource.

core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
rintcius: Happy to change to `hostname/path` but I think we still have to deal with path conflicts.
E.g. suppose there's just a difference in `query` part of URL or even a different header.

core/src/main/scala/quasar/datasource/url/UrlDatasourceConfig.scala
beckyconning: baby steps. if a prospective customer asks for those per url we’ll add that capability.

wemrysi: We could do both, though. i.e. it seems like an arbitrary distinction on the implementation side that compression and format aren't per url while headers are.

However, UX is also important, but we can present a single option for the user and just set the same value for all options.

core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
wemrysi: Yeah, almost makes me wonder if we should just store a name for each resource and have the frontend just prepopulate the field using the heuristics you've defined.

core/src/main/scala/quasar/datasource/url/UrlDatasourceConfig.scala
rintcius: I already changed the server model to just allow multiple URLs.

I think being more flexible in the server than in UI won't work that well because it's now possible to edit datasources too (so then we can't present all valid states properly in UI)

rintcius:
@wemrysi A couple of changes in this PR after your review

- changed format to allow only multiple urls, not headers
- bumped the version in DatasourceType
- changed the way urls are mapped to resource paths (with a different way to ensure the paths are unique)

wemrysi:
We really need to work out how to best deal with config versioning since we've needed it in two places already.

cryogenian:
:+1:

rintcius:
@wemrysi ready for another look. Had to make sure that single URL still work in the old way to avoid having to migrate in frontend.

djspiewak:
@rintcius Does this still need to be **breaking**?

wemrysi:
@rintcius I think @djspiewak is correct that this could now be `feature`. Also, LGTM, feel free to merge whenever it makes sense to do so.

rintcius:
@wemrysi @djspiewak I think it's still breaking because this PR makes sure `url` is auto-upgraded to `urls`. The FE only works with `urls`, not `url` anymore.

djspiewak:
@rintcius Ah that's fair.

PR #271, opened 2019-08-16
Csv/data format

cryogenian:
https://github.com/slamdata/quasar/pull/4305


core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
```diff
@@ -192,13 +161,14 @@ class Http4sUrlDatasourceModule(clientBuilder: Http4sClientBuilder)
...
+  private def singleUrlPath(format: DataFormat, url: Uri): ResourcePath = {
     lazy val defaultLocation: ResourcePath =
-      rt match {
-        case ResourceType.Json => ResourcePath.root() / ResourceName("data.json")
-        case ResourceType.LdJson => ResourcePath.root() / ResourceName("data.ldjson")
+      format match {
+        case DataFormat.Json(JsonVariant.ArrayWrapped, _) => ResourcePath.root() / ResourceName("data.json")
+        case DataFormat.Json(JsonVariant.LineDelimited, _) => ResourcePath.root() / ResourceName("data.ldjson")
+        case _: DataFormat.SeparatedValues => ResourcePath.root() / ResourceName("data.csv")
+        case DataFormat.Compressed(_, _) => ResourcePath.root() / ResourceName("data.gz")
```
wemrysi: I think we should match on the compression scheme here to get compiler help when we add new ones. We can also recurse on the inner format to generate file names like `data.json.gz`.

PR #285, opened 2019-09-04
Misc code cleanup

alissapajer:
Some misc code cleanup in preparation for ch8929


core/src/main/scala/quasar/datasource/url/UrlResource.scala
```diff
@@ -37,21 +37,24 @@ object UrlResource {
       client: Client[F],
       location: ResourcePath,
       req: Request[F])
-      : Resource[F, (Headers, EntityBody[F])] =
+      : Resource[F, EntityBody[F]] =
```
alissapajer: How do we version datasources? Does this change make this patch **breaking**?

alissapajer:
Closing this since I decided that I'll just include these changes with ch8929.

PR #287, opened 2019-09-06
link header pagination

alissapajer:
ch8929


alissapajer:
🛑 is up because I haven't tested this against an actual paginated api yet, and I'd like to do that before merge since datasource unit tests can only test so much.

core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
```diff
@@ -215,14 +228,44 @@ class Http4sUrlDatasourceModule(clientBuilder: Http4sClientBuilder)
...
 
-  private def result[F[_]: Bracket[?[_], Throwable]](
-      urlConfig: UrlDatasourceConfig,
-      urlResource: Resource[F, (Headers, EntityBody[F])])
-      : F[InternalResult[F]] =
-    urlResource.allocated map {
-      case ((_, body), finalize) =>
-        InternalResult(body.onFinalize(finalize), urlConfig.format)
+  @SuppressWarnings(Array("org.wartremover.warts.Recursion"))
+  private def loopRequest[F[_]: ConcurrentEffect: Applicative: ApplicativeError[?[_], Throwable]: MonadResourceErr: Bracket[?[_], Throwable]](
```
alissapajer: @wemrysi extra eyes on this part with regards to resource handling would be much appreciated.

alissapajer:
released as `10.0.4-a73691b`

core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
```diff
@@ -223,14 +228,44 @@ class Http4sUrlDatasourceModule(clientBuilder: Http4sClientBuilder)
...
+        resource.allocated flatMap {
+          case ((headers, body), finalize) =>
+            val nextUri: Option[Uri] = headers collectFirst {
+              case Link(l) if l.rel.exists(_ === LinkHeaderRelation.name.reverseGet(rel)) =>
+	        l.uri
+            }
+            nextUri match {
+              case Some(uri) =>
+                val nextReq = api.copy(req = req.copy(uri = uri))
+                for {
```
wemrysi: This eagerly requests the entire chain of links before returning the stream (due to sequencing the outer `F[_]` of `F[Stream[F, Byte]]` before concatenating the streams). We only need to tease apart the request/streaming effects for the first request to be able to report any environmental errors in the response. Subsequent pages can be lazily requested:
```scala
body.onFinalize(finalize) ++ Stream.force(loopRequest(client, nextReq, config))
```

core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
```diff
@@ -223,14 +228,44 @@ class Http4sUrlDatasourceModule(clientBuilder: Http4sClientBuilder)
...
+                  prevBody <- ConcurrentEffect[F].delay(body.onFinalize(finalize))
+                  nextBody <- loopRequest(client, nextReq, config)
+                } yield prevBody ++ nextBody
+
+              case None =>
+                ConcurrentEffect[F].delay(body.onFinalize(finalize))
```
wemrysi: Adding the finalizer to the `body` stream is pure and doesn't need to be `delay`ed.

core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
alissapajer: ah, good catch!

djspiewak:
[ch8929]

core/src/main/scala/quasar/datasource/url/LinkHeaderRelation.scala
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+package quasar.datasource.url
+
+import slamdata.Predef._
+
+import monocle.Prism
+import scalaz.{Equal, Show}
+
+sealed trait LinkHeaderRelation extends Product with Serializable
```
djspiewak: Why is this a closed algebra? Are these the only four header fields that are ever used for continuations?

alissapajer: @djspiewak according to the specification by @beckyconning, these are the four we want to support. We are not allowing the user to enter their own header relation name. See the description of ch8930.

Fwiw, it would be easy for the backend to support arbitrary header relation names: the UI would provide a text box for the header and we'd simply look for that particular string.

alissapajer:
Also, thanks for the brackets.

alissapajer:
@wemrysi ready for another look!

keeping 🛑 up until I finish testing it against a real paginated api

core/src/main/scala/quasar/datasource/url/LinkHeaderRelation.scala
alissapajer: @djspiewak also fwiw, `next`, `first`, `last`, `prev` appear to be the standard; I haven't found an example api that doesn't use those. I've just looked at a few though.

alissapajer:
released as `10.0.4-2009c16`

alissapajer:
This is blocked on https://github.com/http4s/http4s/issues/2852. I think that should be a relatively straightforward fix, but I need to chat with the http4s folks to see what they think. Plus even if it were fixed fast, I don't know how easily they can release a new version for us.

We have the option of writing a custom parser in the interim. It would only need to find the value between `"<"` and `">"` with the correct `"rel"`, and thus wouldn't need to be a complete parser for all valid links. I think I can do this pretty easily with a regex, so I'll do that.

alissapajer:
Yeah, I'm not regexing this. Parsing is always harder than it should be. Looks like maybe there exists a jvm link header parsing library? I'll poke around.

alissapajer:
ok this claims to do the trick https://jax-rs.github.io/apidocs/2.1/javax/ws/rs/core/Link.html#valueOf-java.lang.String-

alissapajer:
oh except it can only parse a single url :(

alissapajer:
Got it working with `akka-http`!

How to create a datasource that queries for all of djspiewak's repos (there are four pages of paginated results). If I test with too many pages against github I get rate limited.
```
curl -X POST -vvv -H "Content-Type: application/vnd.slamdata.datasource; version=1" -d '{"name":"github-api-djspiewak-repos", "type":{ "name": "url", "version":1 }, "config":{ "requests": [{"url":"https://api.github.com/users/djspiewak/repos", "method":"GET", "body":"", "linkRelation": "next"}], "format": {"type": "json", "precise": false, "variant": "line-delimited"}}}' "http://localhost:8080/api/datasource"
```

Then all queries against the datasource (schema and results) automatically go through the pagination request loop until there are no more pages! I verified that the data browser and an output table contain the full results from all four pages. Pretty snazzy.

alissapajer:
Released as `10.0.4-eb31380`

alissapajer:
@wemrysi got link parsing working. Could you take another look?

build.sbt
```diff
@@ -30,10 +30,12 @@ lazy val core = project
       * `datasourceQuasarVersion`.
       */
     datasourceDependencies ++= Seq(
+      "com.github.julien-truffaut" %% "monocle-core" % "1.6.0",
       "io.argonaut" %% "argonaut-monocle" % "6.2.3",
       "org.http4s" %% "http4s-blaze-client" % "0.20.0-M5",
       "com.codecommit" %% "shims" % "1.7.0",
-
+      "com.typesafe.akka" %% "akka-http" % "10.1.9",
+      "com.typesafe.akka" %% "akka-stream" % "2.5.25",
```
alissapajer: Unfortunately we can't use `akka-http` as a standalone library; I got a `java.lang.ClassNotFoundException: akka.util.ConstantFun$` without `akka-stream`.

I don't know for sure that there's not something else I could depend on to not have to bring in `akka-stream` though.

core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
```diff
@@ -245,14 +249,25 @@ class Http4sUrlDatasourceModule(clientBuilder: Http4sClientBuilder)
...
-              case Link(l) if l.rel.exists(_ === LinkHeaderRelation.name.reverseGet(rel)) =>
-	        l.uri
+            val nextUri: Option[Option[Uri]] = headers collectFirst {
+              case h if h.is(Link) =>
+                HttpHeader.parse("link", h.value) match {
+                  case HttpHeader.ParsingResult.Ok(AkkaLink(links), errors) =>
+                    val relation = LinkParams.rel(LinkHeaderRelation.name.reverseGet(rel))
+                    links collectFirst {
+                      case LinkValue(uri, params) if params.contains(relation) =>
+                        Uri.unsafeFromString(uri.toString)
```
alissapajer: Oh I forgot about this line. We could do the safe thing, and then return `None` if it fails.

alissapajer: Though if it does fail right here, that means that either http4s's parsing or akka's rendering of the uri is buggy, and maybe we don't want to swallow that error?

Semantically, returning `None` means that we won't request any more pages, and so the user could unexpectedly not receive all the pages, even if there are more.

wemrysi: I agree that we want to propagate the error if this happens. Instead of throwing, I think it'd be safer to raise the exception via `MonadError` so we don't have to rely on `F[_]`'s `map` implementation catching exceptions.

alissapajer:
released as `10.0.5-c087ea7`

alissapajer:
@wemrysi ready for another look

PR #297, opened 2019-09-16
Include Stream#scope to work around fs2 bug

alissapajer:
[ch9135]


alissapajer:
released as `10.1.1-ae53e8e`

core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
```diff
@@ -312,7 +313,7 @@ class Http4sUrlDatasourceModule(clientBuilder: Http4sClientBuilder)
                   // We only need to tease apart the request/streaming effects for
                   // the first request to be able to report any environmental errors
                   // in the response. Subsequent pages can be lazily requested.
-                  (body.onFinalize(finalize) ++
```
rintcius: Can you add a comment that `scope` is a workaround and we can remove it after fs2 upgrade?

PR #326, opened 2019-09-27
implement PaginationPlate

alissapajer:
[ch9069]

This `Plate` extracts the value(s) at the provided path. If a scalar at the path locus is seen more than once (via multiple rows or compression in the path), the last scalar value seen is preserved.


core/src/main/scala/quasar/datasource/url/pagination/PaginationPlate.scala
```diff
@@ -0,0 +1,166 @@
...
+
+  def paginationValue(): Option[String] = Option(result)
+
+  private val pathLength: Int = config.path.length
+
+  private val pathKey: Array[Int] = new Array[Int](pathLength)
+  private val indices: Array[Int] = new Array[Int](pathLength)
+  private val fields: Array[String] = new Array[String](pathLength)
+
+  config.path.zipWithIndex foreach {
```
alissapajer: We could put this mutation block in `PaginationPlate.apply`, and then pass in `pathKey`, `indices`, and `fields` as arguments to the private `PaginationPlate` constructor for the class.

alissapajer:
`ContinuationPlate` shares the same infrastructure as `PaginationPlate`, so I'd like to factor out that commonality into a class, allowing the two implementations to override the methods they need to implement differently. @djspiewak any performance degradations you can think of that would result from that?

core/src/main/scala/quasar/datasource/url/pagination/PaginationPlate.scala
```diff
@@ -0,0 +1,166 @@
...
+ * If a scalar at the path locus is seen more than once
+ * (via multiple rows or compression in the path), the last
+ * scalar value seen is preserved.
+ */
+final class PaginationPlate private (config: PaginationConfig)
+    extends Plate[Unit] {
+
+  private var result: String = _
+
+  def paginationValue(): Option[String] = Option(result)
```
djspiewak: Should we have a `reset()` function on here? Right now, I think it's assuming that the plate is re-created with each iteration of `Stateful`. Is that a guarantee?

core/src/main/scala/quasar/datasource/url/pagination/PathComponent.scala
```diff
@@ -0,0 +1,28 @@
...
+
+import slamdata.Predef._
+
+sealed trait PathComponent extends Product with Serializable
+
+object PathComponent {
+  final case class Array(idx: Int) extends PathComponent
+  final case class Object(name: String) extends PathComponent
+  case object CompressedArray extends PathComponent
+  case object CompressedObject extends PathComponent
```
djspiewak: When do we have compressed arrays or objects in tectonic? Can't this just be a two-component ADT? Which would in turn allow `pathKeys` to be `Array[Boolean]`.

core/src/main/scala/quasar/datasource/url/pagination/PaginationPlate.scala
```diff
@@ -0,0 +1,166 @@
...
+
+  private val pathLength: Int = config.path.length
+
+  private val pathKey: Array[Int] = new Array[Int](pathLength)
+  private val indices: Array[Int] = new Array[Int](pathLength)
+  private val fields: Array[String] = new Array[String](pathLength)
```
djspiewak: Randomly, is it worth opening up `FastCursor` from mimir rather than replicating the functionality here?

tectonic-version
```diff
@@ -0,0 +1 @@
+9.0.17
```
djspiewak: Should setup the autobump from tectonic as a 🛑 PR dependent on this one.

djspiewak:
> @djspiewak any performance degradations you can think of that would result from that?

As long as it's a `class` rather than a `trait`, there should be no meaningful performance hit.

core/src/main/scala/quasar/datasource/url/pagination/PathComponent.scala
alissapajer: This ADT is representing the path that is selected by the user in the UI in the "special data browser" when they select the location of their pagination value. Object and array flattening is allowed, and needs to be represented herel so we can select the last value seen at a flattened path.

"compressed" is the wrong word to use here though. `FlattenedArray` and `FlattenedObject` would better reflect the intent.

core/src/main/scala/quasar/datasource/url/pagination/PaginationPlate.scala
alissapajer: That's a good question. We have the option of either recreating the plate or resetting. I think that resetting will be the better option so that we don't have to recreate the plate. So I'll add a `reset()` function.

core/src/main/scala/quasar/datasource/url/pagination/PaginationPlate.scala
alissapajer: @djspiewak opinions on this one way or the other? I'm thinking that it's cleaner in `Pagination.apply`, especially since it's part of the "start-up" of the plate, and we only ever want to do it once per plate initialization.

core/src/main/scala/quasar/datasource/url/pagination/PaginationPlate.scala
alissapajer: The path that we use here has flattened components, whereas `FastCursor` doesn't. 

Also, I need to be able to determine the path component at a particular index; does `FastCursor` allow for that?

PR #328, opened 2019-09-30
implement ContinuationPlate

alissapajer:
depends on https://github.com/slamdata/quasar-datasource-url/pull/326

[ch9361]


alissapajer:
@djspiewak this implementation is very similar in structure to #326, so any changes requested there will likely have to be made here as well. I also will likely want to factor out the commonalities in the implementations if we can do so without a performance overhead.

core/src/main/scala/quasar/datasource/url/pagination/ContinuationPlate.scala
```diff
@@ -0,0 +1,275 @@
...
+  private def atFork: Boolean = depthBelowFork == 0
+  private def atScalar: Boolean = forkPoint == pathLength
+}
+
+object ContinuationPlate {
+  def apply[F[_]: Sync](config: ContinuationConfig): F[ContinuationPlate] =
+    Sync[F].delay(new ContinuationPlate(config))
+}
+
+object NumericConversions {
```
alissapajer: This is here temporarily and will be replaced with https://github.com/slamdata/quasar/pull/4379

alissapajer:
@djspiewak I pushed all the changes for #326  to this PR so that we can merge this one and have both features land at once.

Let me know if there are more changes you wanted to see; otherwise this is good to go.

I didn't include the abstraction that will generalize both plates. I'm working on that now. This PR can be merged with or without that.

alissapajer:
@djspiewak push `CursorPlate`. This is ready now.

core/src/main/scala/quasar/datasource/url/pagination/ContinuationPlate.scala
alissapajer: Replaced.

PR #357, opened 2019-10-14
support data pagination

alissapajer:
[ch9071]


alissapajer:
released as `12.0.7-1158c28`

@rintcius snapshot published so you can test the frontend work against it!

alissapajer:
Tested this against the twitter api with the following config:
```

curl -X POST -vvv -H "Content-Type: application/vnd.slamdata.datasource; version=1" -d '{"name":"list-members-2", "type":{ "name": "url", "version":1 }, "config":{ "requests": [{"url":"https://api.twitter.com/1.1/lists/members.json?count=30&list_id=111244779", "method":"GET", "body":"", "pagination": { "dataPagination": {"nextRequest": {"path":["next_cursor"], "type": "Number", "url": "https://api.twitter.com/1.1/lists/members.json?count=30&list_id=111244779&cursor={{reform_pagination_value}}"}, "requestCondition": {"path": ["next_cursor"], "condition": "StopWhen", "value": 0}}}}], "format": {"type": "json", "precise": false, "variant": "line-delimited"}, "headers": {"authorization": "Bearer <REDACTED>"}}}' "http://localhost:8080/api/datasource"
```

To replicate this exactly, you'll need to register a twitter application and generate a bearer token. The twitter list queried is https://twitter.com/ches/lists/fp.

core/src/test/scala/quasar/datasource/url/UrlDatasourceConfigSpec.scala
```diff
@@ -507,7 +629,412 @@ class UrlDatasourceConfigSpec extends Specification {
...
+            |    {
+            |      "url": "https://test.com",
+            |      "method": "POST",
+            |      "body": "{\\"id\\": 42}"
+            |    }
+            |  ]
+            |}
+          """.stripMargin
+
+        s.decodeOption[UrlDatasourceConfig] must_=== None
```
rintcius: This was already there, but testing failures with `None` can easily test the wrong thing, if there's some unintended error in the json, especially now the json has become more complex.
Maybe time to start verify that the error contains some string instead?

alissapajer: 👍 some of the other failure tests already do this. I'll audit all the failure tests in this spec.

core/src/main/scala/quasar/datasource/url/pagination/ContinuationType.scala
```diff
@@ -18,9 +18,40 @@ package quasar.datasource.url.pagination
...
+    Prism[String, ContinuationType] {
+      case "ContinueWhile" => Some(ContinueWhile)
+      case "StopWhen" => Some(StopWhen)
+      case _ => None
+    } {
+      case ContinueWhile => "ContinueWhile"
+      case StopWhen => "StopWhen"
+    }
+
+  implicit val codecContinuationType: CodecJson[ContinuationType] =
```
alissapajer: I put the codecs for these algebras in the companion object to encourage there being only one codec, and to make them accessible.

The unit tests for these are still all in `UrlDatasourceConfigSpec` though. I'd like to move the tests for `ContinuationType` into a `ContinuationTypesSpec`, etc. But before taking the time to do that, I wanted to see if there was any reason not to do that. In sdbe we haven't always organized things this way, and I wasn't sure if it was for any particular reason.

core/src/main/scala/quasar/datasource/url/ApiRequest.scala
```diff
@@ -31,6 +35,24 @@ object ApiRequest {
...
+      req: UriRequest,
+      pagination: PaginationConfig,
+      nextPage: String,
+      continuation: Option[ContinuationConfig])
+      extends ApiRequest {
+
+    def nextPageUri(data: String): Option[UriRequest] = {
+      import UrlDatasourceConfig._
+
+      val token = "\\{\\{reform_pagination_value\\}\\}".r
```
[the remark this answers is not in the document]
alissapajer: Is there an equivalent of `scala.util.matching.Regex.replaceAllIn` using the java `Pattern`?

Since this is a straightforward regex pattern, it just needs a few unit tests (which are included), so I'm not sure I see the benefit of `Pattern.quote`.

alissapajer:
Updated the README doc

core/src/test/scala/quasar/datasource/url/UrlDatasourceConfigSpec.scala
alissapajer: For the record, I misremembered, and it's the `Http4sUrlDatasourceSpec` that currently checks for specific errors.

core/src/main/scala/quasar/datasource/url/ApiRequest.scala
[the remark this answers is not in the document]
rintcius: Oh actually there's a `quote` in scala Regex too. Can you use: 

```
Regex.quote("{{reform_pagination_value}}").r
```

?

core/src/test/scala/quasar/datasource/url/UrlDatasourceConfigSpec.scala
alissapajer: Pushed up a change to verify that the decode failures are indeed decode failures and not parse failures. We aren't retaining error messages though, so I'm going to look into that to see if we can hold on to any more information when we fail to decode.

core/src/main/scala/quasar/datasource/url/ApiRequest.scala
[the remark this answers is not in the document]
alissapajer: 👍 done.

README.md
```diff
@@ -45,11 +45,97 @@ A request has the following format.
...
+
+#### `"path"`
+
+Each element in the `"path"` array is one of
+- `String`: field name
+- `Number`: array index
+- `true`: flattened object
+- `false`: flattened array
+
+For example, if the user selects `All Keys -> All Indices -> key "foo" -> index 2`, that corresponds to the path `[true, false, "foo", 2]`. We will not allow the user to select a field name or an array index after selecting an `All keys` or `All indices`, respectively.
```
wemrysi: What is the behavior if a path containing an "All Keys"/"All Indices" produces multiple values?

If we think a human will ever specify this field, a more memorable syntax might be nice. Maybe something like `CPath`'s, e.g. `{}[].foo[2]`. Probably not necessary now.

core/src/main/scala/quasar/datasource/url/pagination/ContinuationType.scala
wemrysi: Sounds good to me :+1:. In SDBE a lot of the codecs are for types not introduced within SDBE itself (mostly from `quasar.api.*`), so there usually isn't companion to place them in.

README.md
```diff
@@ -45,11 +45,97 @@ A request has the following format.
...
+An optional `"pagination"` is included in the api connector config. When this field is present, it indicates that the datasource is paginated. When this field is absent, the datasource is not paginated.
+
+There are two types of pagination: data-pagination and header-pagination. If the datasource is data-paginated, a `"dataPagination"` field will be present below `"pagination"`. If the datasource is header-paginated, a `"headerPagination"`field will be present below `"pagination"`. It is not valid for a datasource to be both header-paginated and data-paginated.
+
+```json
+{
+  "url": String,
+  "method": "GET" | "HEAD" | "POST" | "PUT" | "CONNECT" | "OPTIONS" | "TRACE" | "PATCH",
+  "body": String,
+  "pagination": { "headerPagination": ... } | { "offsetPagination": ... } | { "dataPagination": ... }
```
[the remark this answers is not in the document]
alissapajer: Oh, I'll remove that. I copied this from the config I'm writing up for offset pagination, which is the next feature I'm adding :)

README.md
alissapajer: The last value seen at that flattened path will be used. This behavior is tested in the unit tests `PaginationPlate` and `ContinuationPlate`. I'll add a line specifying this behavior here.

Currently a human specifies this indirectly via the UI by picking a path in the data browser (really a custom data browser meant for this purpose, which @rintcius is working on). So a human will never have to write this. 

Also iirc, when I was working on the json encoding for `scion` diffs, @garyb said that empty object and empty array are harder (slower?) to handle in the frontend, so we opted for `true` and `false` to denote object/array flattening. So I stuck with the same here.

README.md
```diff
@@ -45,11 +45,97 @@ A request has the following format.
...
+
+Each element in the `"path"` array is one of
+- `String`: field name
+- `Number`: array index
+- `true`: flattened object
+- `false`: flattened array
+
+For example, if the user selects `All Keys -> All Indices -> key "foo" -> index 2`, that corresponds to the path `[true, false, "foo", 2]`. We will not allow the user to select a field name or an array index after selecting an `All keys` or `All indices`, respectively.
+
+#### `"type"`
```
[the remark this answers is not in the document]
alissapajer: In the UI design, the user specifies if the value is an integer or a string, exclusive. @rintcius could you confirm that this is correct?

So although the type doesn't matter in the implementation, this is required to implement the feature as spec'd. We wanted to avoid any confusion in the case when the same field contains a value of a different type than expected.

README.md
wemrysi: Sounds good. Regarding human input, I meant to specify that we'd parse that format from a `String` but forgot the quotes.

garyb: Yeah, empty object/array incur a construction cost that `true` / `false` don't suffer from, as the latter are constant values rather than things that need prototypal instantiation.

alissapajer: @garyb for reference, would strings like `"{}"` and `"[]"` incur a similar overhead as object/array construction, or could you treat them like constants?

README.md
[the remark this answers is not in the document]
rintcius: @alissapajer What do you mean exactly with the type doesn't matter in the implementation? My expectation is that the user gets different behaviour when selecting `foo (String)` vs selecting `foo (Number)`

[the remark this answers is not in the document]
alissapajer: @rintcius you are correct that the user gets different behavior when selecting `foo (String)` vs selecting `foo (Number)`. 

>So although the type doesn't matter in the implementation

What I said there was a stronger statement than what I meant. What I meant was "the type isn't relevant when we turn the value into a string and insert it into the url".

PR #386, opened 2019-11-20
implement offset pagination

alissapajer:
[ch9986]
[ch9987]
[ch9973]
[ch10145]


alissapajer:
released as `15.0.0-ac11cea`

@cryogenian you can base your variable work against this PR

alissapajer:
example datasource:
```
curl -X POST -vvv -H "Content-Type: application/vnd.slamdata.datasource; version=1" -d '{"name":"openfda-osteo", "type":{ "name": "url", "version":1 }, "config":{ "requests": [{"offsetPaginated":{"perRequest":100,"request":{"url":"https://api.fda.gov/drug/event.json?limit={{reform_per_request}}&skip={{reform_offset}}&search=patient.drug.drugindication.exact:OSTEOARTHRITIS&api_key=<REDACTED>","method":"GET","body":""},"stopCondition":{"totalExceeded":{"eachRequest":{"path":["meta","results","total"]}}}}}], "format": {"type": "json", "precise": false, "variant": "line-delimited"}}}' "http://localhost:8080/api/datasource"
```

core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
```diff
@@ -259,72 +264,91 @@ class Http4sUrlDatasourceModule(clientBuilder: Http4sClientBuilder)
...
-            case Some(s) =>
-              dp.nextPageUri(s) match {
-                case Some(next) => makeBasicRequest(client, next, config)
-                case None => Sync[F].pure(Stream.empty)
-              }
+            case Some((offset, total)) =>
+              if (offset >= total)
+                Stream.empty
+              else
+                getData(client, req, config, pag.nextPageUri(_))(Some(offset.toString))
```
alissapajer: The recursive component (i.e. do we request another page, or do we stop) of the pagination logic here (and in all the cases in this function) is not unit tested. I'm not sure exactly how to test it. (It's also not tested for token pagination.)

In `Http4sUrlDatasourceModuleSpec` we currently test a single call to `Stateful#data` (for each of offset and token pagination), but we don't test multiple calls to it, as will happen in actual datasources. In order to test this, we'd need to mock a `Stateful` evaluator that gets the state and then gets the data, and loops this until it's done.

alissapajer: This will allow us to test:
1) we actually stop when the offset reaches the total
2) we request each page exactly once

To be clear, we have many unit tests which test that the various `Plate`s extract the correct information (stop conditions, tokens, total, etc.). We just don't test the very final step of doing the end-to-end recursion with `Stateful`.

core/src/test/scala/quasar/datasource/url/pagination/CursorPlateSpec.scala
```diff
@@ -0,0 +1,935 @@
...
+        plate.unnest()
+        plate.atTarget must beFalse
+        plate.unnest()
+        plate.atTarget must beFalse
+        plate.nestArr()
+        plate.atTarget must beFalse
+        plate.nestMap("a")
+        plate.atTarget must beFalse
+        plate.nestArr()
+        plate.atTarget must beTrue // true
```
rintcius: looks like a couple of unnecessary comments

core/src/main/scala/quasar/datasource/url/ApiRequest.scala
```diff
@@ -40,19 +37,45 @@ object ApiRequest {
...
+      stopCondition: Either[ValueEquals, TotalExceeded])
+      extends ApiRequest {
+
+    def nextPageUri(offset: String): Option[UriRequest] = {
+      import UrlDatasourceConfig._
+
+      val replaced: String =
+        replaceReformPerRequest(
+          replaceReformOffset(nextPage, offset),
+          perRequest.toString)
```
rintcius: Maybe for later... wondering if we should support vars in the body as well

core/src/test/scala/quasar/datasource/url/pagination/CursorPlateSpec.scala
alissapajer: Actually I put those comments in on purpose to point out where the `true` values are. Since the test body is long, I wanted the important parts to stand out.

core/src/main/scala/quasar/datasource/url/ApiRequest.scala
alissapajer: 👍 if we come across a web api that uses the request body to provide pagination information, then we could add this as a feature in the future.

PR #394, opened 2019-12-04
Variables/encoding

cryogenian:
[ch10229] [ch10259]

+ Extracted codecs and moved them to their respective companion objects
+ Moved from scalaz to cats in pagination package
+ Implemented idempotent uri encoder
+ Integrated the encoder with `Substitution` and `ApiRequest`


core/src/main/scala/quasar/datasource/url/UriEncoded.scala
```diff
@@ -0,0 +1,167 @@
...
+
+  def apply(nonEncoded: String): Option[UriEncoded] = {
+    val inp: List[Char] = nonEncoded.toList
+    val out: CharBuffer = CharBuffer.allocate(nonEncoded.length * 3)
+    val result = collectChars(inp, out, Scheme(""))
+    result.map(new UriEncoded(_))
+  }
+
+  @scala.annotation.tailrec
+  private def collectChars(
```
rintcius: What's the exact reason we need the state machine here? I was thinking we could get away with some string replacements on top of the standard URL encoder

cryogenian: Right, if we don't care about `[]` inside ipv6 host we can do something like `replaceDoubleEncoded(Regex.quote("+").r.replaceAllIn(URLEncoder.encode(x, StandardCharsets.UTF_8.toString), "%20"))` where `replaceDoubleEncoded` is something like 
`Regex.quote("%25([0-9A-F]{2})").r.replaceAllIn(_, "%$1"))`. 

Actually I just realized that I need to check that it's `[ipv6]` not `[anything]` :)

core/src/test/scala/quasar/datasource/url/UriEncodedSpec.scala
```diff
@@ -19,60 +19,57 @@ package quasar.datasource.url
...
+      Uri.fromString(UriEncoded(inp).value) must beRight
     }
   }
+
+  "property" >> {
+    "encoding is idempotent" >> prop { (x: String) =>
+      val inp = s"http://$x"
+      UriEncoded(UriEncoded(inp).value).value must_=== UriEncoded(inp).value
+    }.set(minTestsOk = 10000).setArbitrary(Arbitrary(Gen.asciiStr))
+  }
```
rintcius: Nice! 

I guess we could improve the generator, but feel free to leave like this

core/src/main/scala/quasar/datasource/url/UriEncoded.scala
```diff
@@ -19,149 +19,57 @@ package quasar.datasource.url
...
-      else if (state.startCollecting(x))
-        collectChars(tail, accum, state.collect(x))
-      else if (state.omitted(x))
-        collectChars(tail, accum.put(x), state.transit(x))
-      else if (state.valid(x))
-        collectChars(tail, accum.put(mbEncode(x)), state.clear)
-      else
-        None
-  }
+  private def quoteDollar(inp: String): String =
```
rintcius: did you mean `escapeDollar`?

PR #397, opened 2019-12-05
rate limit

alissapajer:
[ch10224]


alissapajer:
@djspiewak I've been staring at this thinking about if the effect is actually threaded through the correct places; I decided to PR it so we can look at it as a whole and figure it out.

core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
```diff
@@ -186,11 +200,19 @@ class Http4sUrlDatasourceModule(clientBuilder: Http4sClientBuilder)
...
+      token <- Resource.liftF(SingletonToken[F]())
       pr <- Resource.liftF(Sync[F].delay(java.util.UUID.randomUUID().toString))
       mkParser = (x: List[Substitution]) => new TemplatedUri.Parser(pr, x)
       ds <- result(mkParser) match {
         case Left(a) => Resource.pure[F, Either[InitializationError[Json], DS[F]]](Left(a))
-        case Right(cfg) => mkDatasource(cfg, sanitizedConfig)
+        case Right(cfg) =>
+          val limit: F[F[Unit]] = cfg.rateLimit match {
+            case Some(rl) =>
+              rateLimiter(token, rl.max, new FiniteDuration(rl.window, TimeUnit.MILLISECONDS))
```
djspiewak: BTW, `rl.window.millis` also works :-)

PR #407, opened 2019-12-17
reset state in token paginated token present

alissapajer:
[ch10364]


alissapajer:
released as `16.0.2-5b50c8a`

PR #431, opened 2020-01-14
Support optional OAuth2 authentication

wemrysi:
Adds support for optional OAuth2 authentication via the `authorization_code` flow.

[ch9894]
[ch9895]


wemrysi:
Just realized I forgot to update the docs, will do so later today.

alissapajer:
@wemrysi if an oauth token expires in the middle of running a paginated query, will the token refresh happen asynchronously to the query and additionally not interrupt the query? Or would the token expiration not be detected mid-query, meaning that we'd terminate the running query prematurely?

core/src/main/scala/quasar/datasource/url/oauth/OAuth2GrantConfig.scala
```diff
@@ -0,0 +1,73 @@
...
+      scope: Option[Scope],
+      clientId: ClientId,
+      clientSecret: ClientSecret,
+      clientAuthnMethod: ClientAuthenticationMethod,
+      redirectUri: Option[Uri])
+      extends OAuth2GrantConfig {
+
+    def sanitized = copy(
+      authorizationCode = AuthorizationCode("<REDACTED>"),
+      clientSecret = ClientSecret("<REDACTED>"))
```
alissapajer: There is a `RedactedStr` [here](https://github.com/slamdata/quasar-datasource-url/blob/master/core/src/main/scala/quasar/datasource/url/package.scala#L22) - could you use that?

wemrysi:
@alissapajer Unless the source proactively terminates an in-process response stream due to an access token invalidation, refreshing the token in the middle of pagination should happen transparently.

alissapajer:
@wemrysi this implementation gives one set of OAuth credentials per datasource, but each datasource can contain multiple requests. Do we need to generalize this in the future so that we can configure the OAuth credentials per request url rather than per datasource?

wemrysi:
@alissapajer I'm not sure, discussions thus far have assumed a single set of credentials, which also seems very common in the use cases (OAuth domain is the same as the api domain). If one was creating a single url datasource that talked to multiple root domains, then it'd probably be necessary. Shouldn't be terribly difficult as I think we'd just change the config and where we wrap the underlying `Client` in `OAuth2Middleware`.

alissapajer:
@wemrysi there are a few conflicts now

PR #466, opened 2020-02-07
Improve access token response handling

wemrysi:
Fixes an issue when decoding the `scope` of a successful access token response and sanitizes any error messages to not contain the response text to avoid leaking credentials.


core/src/main/scala/quasar/datasource/url/oauth/types.scala
```diff
@@ -85,21 +84,16 @@ object ClientSecret {
...
 
-final case class ScopeToken(value: String) extends AnyVal
-
-object ScopeToken {
-  implicit val codecJson: CodecJson[ScopeToken] =
-    CodecJson.derived[String].xmap(ScopeToken(_))(_.value)
-}
-
-final case class Scope(tokens: NonEmptyList[ScopeToken]) extends AnyVal
+final case class Scope(tokens: NonEmptyList[String]) extends AnyVal
```
rintcius: Ok so the no scope scenario mentioned here https://developer.github.com/apps/building-oauth-apps/understanding-scopes-for-oauth-apps/#available-scopes would be handled by not supplying the scope field, right?

README.md
```diff
@@ -92,7 +92,7 @@ Authentication via the OAuth 2.0 authorization code flow may be enabled by inclu
     "clientId": String,
     "clientSecret": String,
     "clientAuthenticationMethod": "basic" | "request-body"
-    [, "scope": [String]]
+    [, "scope": String]
```
rintcius: Ah so you changed the scope from array to string to match the RFC, right?

wemrysi: Yes, that seemed more straightforward.

core/src/main/scala/quasar/datasource/url/oauth/types.scala
wemrysi: Correct :+1:

PR #469, opened 2020-02-12
rate limiting as middleware

alissapajer:


core/src/main/scala/quasar/datasource/url/RateLimitingMiddleware.scala
```diff
@@ -0,0 +1,49 @@
...
+import org.http4s.client.Client
+
+object RateLimitingMiddleware {
+
+  def apply[F[_]: Sync](
+      rl: RateLimiterEffects[F],
+      client: Client[F])
+      : Client[F] = {
+
+    def catch429(req: Request[F]): Resource[F, Response[F]] =
```
wemrysi: Not quite as good as catch-22, but still decent 😉

alissapajer: I had that thought as well! :)

PR #472, opened 2020-02-14
minor splitting of a function

alissapajer:


core/src/main/scala/quasar/datasource/url/ApiRequest.scala
```diff
@@ -48,24 +48,26 @@ object ApiRequest {
...
-  def nextPageUri(apiRequest: ApiRequest[_], data: String)
-      : Option[UriRequest[Uri]] = apiRequest match {
-    case TokenPaginated(req, _, nextPage, _) =>
-      Uri.fromString(UriEncoded(replaceReformToken(nextPage, data)).value).toOption.map(x => req.copy(uri = x))
-    case OffsetPaginated(req, perRequest, nextPage, nextBody, _) =>
-      val doReplace = (inp: String) =>
-        replaceReformPerRequest(replaceReformOffset(inp, data), perRequest.toString)
-      Uri.fromString(UriEncoded(doReplace(nextPage)).value).toOption.map(x => req.copy(uri = x, body = doReplace(nextBody)))
-    case _ =>
-      None
```
alissapajer: Previously we were using `None` to signify both an invalid `Uri` and an unexpected type of `ApiRequest`. We can improve this by writing one function per request type.

PR #475, opened 2020-02-21
pass through headers and format individually

alissapajer:
also use `Request#withEntity` to compute the content length


core/src/test/scala/quasar/datasource/url/UrlResourceSpec.scala
```diff
@@ -0,0 +1,76 @@
...
+import cats.Id
+
+object UrlResourceSpec extends Specification {
+
+  "set content length" >> {
+    "empty body" in {
+      val req = UriRequest(Uri.uri("www.foo.com"), Method.GET, "")
+
+      UrlResource.request[Id](req, Map.empty).headers mustEqual(
+        Headers.of(Header("Content-Length", "0"), Header("Content-Type", "text/plain; charset=UTF-8")))
```
alissapajer: We now include the `"Content-Type"` header by default. We could version this as **feature**?

core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
```diff
@@ -279,24 +282,25 @@ class Http4sUrlDatasourceModule(clientBuilder: Http4sClientBuilder)
   private def makeResult[F[_]: ConcurrentEffect: MonadResourceErr](
       client: Client[F],
       api: ApiRequest[Uri],
-      config: Config[Uri])
+      headers: Map[CIS, String],
+      format: DataFormat)
```
alissapajer: Passing around the full config here was making the initial request feature confusing to implement because it wasn't clear by the types which component of the config was actually being used. So that's what motivated this change now.

PR #486, opened 2020-03-05
Extend valueEquals stop condition with support for empty containers

wemrysi:
[ch10903]


rintcius:
Just needs a merge from master

PR #502, opened 2020-04-07
Update quasar to 168.0.2

wemrysi:


core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
```diff
@@ -591,21 +599,20 @@ class Http4sUrlDatasourceModule(clientBuilder: Http4sClientBuilder)
...
             case Right(Some(uri)) =>
               val nextReq = api.copy(req = api.req.copy(uri = uri))
+              val nextBytes = loopHeaderPaginated(client, nextReq, configHeaders, format)
 
               // We only need to tease apart the request/streaming effects for
               // the first request to be able to report any environmental errors
               // in the response. Subsequent pages can be lazily requested.
-              (body.onFinalize(finalize) ++
-                Stream.force(loopHeaderPaginated(client, nextReq, configHeaders, format))).pure[F]
+              (body ++ Stream.resource(nextBytes).flatten).pure[F]
```
alissapajer, line 615: Confirming that `++` is creating a new scope from the perspective of the resource finalizers: meaning that we still have the effect of finalizing the `body` before the the `nextBytes` are streamed?

wemrysi: Hmmm, so `body` will be finalized prior to `nextBytes`, however the outer `Resource` won't be, which is undesirable. I'm not sure how to achieve eager finalization of both without reverting to what we were doing before (which has the downside of being possible to never finalize the `Resource` if the stream isn't consumed. That is unsatisfying :disappointed:

wemrysi: I guess we could use `.allocated` and register the finalizer on both the `body` and as the `release` of a new resource, with an `MVar` or other sort of latch that ensures it is only run once.

core/src/test/scala/quasar/datasource/url/Http4sUrlDatasourceModuleSpec.scala
```diff
@@ -2052,19 +1723,38 @@ object Http4sUrlDatasourceModuleSpec extends Specification {
...
+        case _ =>
+          IO.pure(List())
+      }
+    }
+
+  def evaluateTyped(dsmod: Http4sUrlDatasourceModule, cfg: Json, name: String)
+      : IO[Either[IE, Either[String, Json]]] =
+    evaluateTyped(dsmod, cfg, ResourcePath.root() / ResourceName(name))
+
+  def evaluateTyped(dsmod: Http4sUrlDatasourceModule, cfg: Json, path: ResourcePath)
```
alissapajer, line 1748: nice refactoring :) :)

alissapajer:
@wemrysi approved, pending the manual testing of a header-paginated source, which I think is a good idea. While your running the manual test, could you monitor the connections to the api endpoint to confirm that all resources are released when the query completes?

wemrysi:
@alissapajer yes, things are a bit weird as everything works nicely when the `/**` resource is requested but the connection is closed before the end of the stream when requesting a single resource directly. Am looking into it.

alissapajer:
@wemrysi gotcha. I'm going to unapprove as a reminder that there are some outstanding issues.

alissapajer:
Also, unit testing this connector is hard :/ If you have any ideas about how we could catch whatever the problem turns out to be in a unit test, we should evaluate how much work that test would be to write.

wemrysi:
@alissapajer ready for another look, turns out the root cause of the issues I was seeing was due to how `Resource`s were being used in sdbe, not here.

PR #509, opened 2020-04-21
Added request token authentication

beckyconning:
Not sure what is required in regards to testing. @alissapajer please let me know : )

This also subsumes client credentials oauth grant.


alissapajer:
@beckyconning is there a clubhouse issue for this?

core/src/main/scala/quasar/datasource/url/reqTokenAuth/ReqTokenAuthMiddleware.scala
```diff
@@ -0,0 +1,166 @@
...
+import org.http4s.util.{CaseInsensitiveString => CIS}
+import org.typelevel.jawn.{AsyncParser, ParseException}
+import quasar.api.resource.ResourcePath
+import quasar.connector.{ByteStore, MonadResourceErr, ResourceError}
+import quasar.datasource.url.pagination.ValueType
+import quasar.datasource.url.{GatheredInfo}
+import scodec.Codec
+import scodec.bits.BitVector
+
+object ReqTokenAuthMiddleware {
```
alissapajer, line 39: To test this, can you add tests similar to the ones in `OAuth2MiddlewareSpec`. Looks like most if not all of the test cases in that spec will apply to this situation as well.

core/src/main/scala/quasar/datasource/url/reqTokenAuth/ReqTokenAuthMiddleware.scala
```diff
@@ -0,0 +1,166 @@
...
+  val InfoMaxSizeBytes = 2000000
```
alissapajer, line 40: Use the one [here](https://github.com/precog/quasar-datasource-url/blob/master/core/src/main/scala/quasar/datasource/url/package.scala#L42)

core/src/main/scala/quasar/datasource/url/reqTokenAuth/ReqTokenAuthMiddleware.scala
```diff
@@ -0,0 +1,166 @@
...
+              res.pure[Resource[F, ?]]
+          }
+
+        case None =>
+          attemptRequest(req)
+      }
+
+
+    Client[F](attemptCached)
+  }
```
alissapajer, line 84: Could we factor out the similarities between this and `OAuth2Middleware.apply` into a `HttpAuthenticationMiddleware`. The `apply` function above and [here](https://github.com/precog/quasar-datasource-url/blob/master/core/src/main/scala/quasar/datasource/url/oauth/OAuth2Middleware.scala#L40) are identical modulo this bit
```
authn <- Resource.liftF(authorizationCode(store, config, client))
```
so we could have a class that takes the `Resource[F, HttpAuthentication]`:
```scala
 def apply[F[_]: MonadResourceErr: Sync](
      store: ByteStore[F],
      httpAuth: Resource[F, HttpAuthentication])(
      client: Client[F])
      : Client[F] = {

    def accessDenied[A](detail: String, cause: Option[Throwable]): F[A] = ... // duplicate stuff

    def attemptRequest(req: Request[F]): Resource[F, Response[F]] =
      for {
        authn <- httpAuth
        initialRes <- client.run(HttpAuthentication(authn, req))
        res <- Resource liftF { ... } // duplicate stuff
      } yield res

    def attemptCached(req: Request[F]): Resource[F, Response[F]] = ... // duplicate stuff
    
    Client[F](attemptCached)
  }
```
and then we'd have a single notion of a `Client` that checks the cache for authentication, and makes a new request if it the cache is empty or the auth fails.

That will make it easier to add another one of these in the future, as well as making it more maintainable.

core/src/main/scala/quasar/datasource/url/reqTokenAuth/ReqTokenAuthMiddleware.scala
```diff
@@ -0,0 +1,166 @@
...
+        .flatMap(_.body)
+        .through(GatheredInfo.limitSize(InfoMaxSizeBytes, limitExceeded))
+        .through(parseJson)
+        .flatMap(extractValues _)
+        .take(1)
+        .compile.toList.flatMap(_.headOption.fold(noToken)(x => HttpAuthentication.bearerToken(x).pure[F]))
+
+  }
+
+  private[reqTokenAuth] implicit class HttpAuthStoreOps[F[_]](store: ByteStore[F]) {
```
alissapajer, line 150: We should also de-dup this from here and `OAuth2Middleware`.

core/src/main/scala/quasar/datasource/url/reqTokenAuth/ReqTokenAuthMiddleware.scala
```diff
@@ -0,0 +1,166 @@
...
+    client: Client[F])
+    : F[HttpAuthentication] = {
+
+      def addContentLength(body: String, headers: Map[CIS, String]): Map[CIS, String] = {
+        val contentLength = body.getBytes(StandardCharsets.UTF_8).length
+        headers + (CIS("Content-Length") -> contentLength.toString)
+      }
+
+      val hdrs = addContentLength(config.req.body, config.headers)
+        .filterNot(_._1.equals(CIS("Content-Length")))
```
alissapajer, line 100: Are you adding the `"Content-Length"` header and then removing it?

alissapajer: `withEntity`, which you're using below on line 108 will include the `"Content-Length"` header automatically, so I wouldn't expect you to have to manually include it.

https://github.com/http4s/http4s/blob/master/core/src/main/scala/org/http4s/Message.scala#L65

alissapajer: Ooo maybe the goal is to overwrite the `"Content-Length"` header in the case when the user provides one.

core/src/main/scala/quasar/datasource/url/reqTokenAuth/ReqTokenAuthConfig.scala
```diff
@@ -0,0 +1,49 @@
...
+object ReqTokenAuthConfig {
+
+  private implicit val codecHeaders: CodecJson[Map[CIS, String]] =
+    CodecJson(
+      (hs: Map[CIS, String]) => hs.toList.map(_.leftMap(_.value)).toMap.asJson,
+      c => c.as[Map[String, String]].map(_.toList.map(_.leftMap(CIS(_))).toMap))
+
+  implicit def eqReqTokenAuthConfig: Eq[ReqTokenAuthConfig] = Eq.fromUniversalEquals
+
+  implicit def showReqTokenAuthConfig: Show[ReqTokenAuthConfig] = Show.fromToString
```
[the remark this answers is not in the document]
beckyconning: could you provide this as a suggestion i can apply as i'm not sure how this fits in.

core/src/main/scala/quasar/datasource/url/reqTokenAuth/ReqTokenAuthMiddleware.scala
beckyconning: yes that is the intention. weirdly the content length heading doesn't seem to be removed last time i checked. none the less yes this filtering should be done before we add it.

alissapajer: Ok, so we can omit `addContentLength` since `withEntity` is going to include the correct `"Content-Length"` header for us. We want the user-defined headers to take precedence in all cases except for the case of `"Content-Length"`. 

To acheive this, we can implement `req` below with `UrlResource.request`. That will do the right thing with the headers.

alissapajer: Also could you add a comment to `UrlResource.request` that says
```
the user-defined headers to take precedence in all cases except for the case of "Content-Length"
```

beckyconning: > withEntity is going to include the correct "Content-Length" header for us

my perception was that i fixed an issue i had connecting to a specific api and finding that my content length header was missing or wrong by adding this content length code. it might have been a misperception though.

beckyconning: oh it might have been because i was using `withHeaders` rather than `putHeaders` at some point? i'm not sure.

core/src/main/scala/quasar/datasource/url/reqTokenAuth/ReqTokenAuthMiddleware.scala
```diff
@@ -0,0 +1,166 @@
...
+            None)))
+
+      val noToken =
+        MonadResourceErr[F].raiseError[HttpAuthentication](
+          ResourceError.connectionFailed(
+            ResourcePath.root(),
+            Some("Couldn't find any tokens at specified path"),
+            None))
+
+      client.stream(req)
```
alissapajer, line 140: There's not really a reason to contain the lifecycle of the request in a singleton `Stream` and then call `take(1)` below. 

Instead you can use [`Client#run`](https://github.com/http4s/http4s/blob/master/client/src/main/scala/org/http4s/client/Client.scala#L17.) This will lift the response into a `Resource[F, Response[F]]`, and then `authorizationCode` can return `Resource[F, HttpAuthentication]`.

core/src/main/scala/quasar/datasource/url/reqTokenAuth/ReqTokenAuthMiddleware.scala
alissapajer: I'd try doing `val req = UrlResource.request(config.req, config.headers)`, and deleting `def addContentLength` and `val hdrs`. That's what we do elsewhere in the url connector so I'd expect that to work. If it doesn't it probably means we have bugs elsewhere as well :)

core/src/main/scala/quasar/datasource/url/reqTokenAuth/ReqTokenAuthConfig.scala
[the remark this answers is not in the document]
alissapajer:
```scala
implicit def showReqTokenAuthConfig: Show[ReqTokenAuthConfig] =
  Show show {
    case ReqTokenAuthConfig(req, path, headers) =>
      s"ReqTokenAuthConfig(${req.show}, ${path.show}, ${headers.show})"
  }
```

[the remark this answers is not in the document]
alissapajer: In general I aim to just use `Show.fromToString` only for `case object`s and then use custom `Show` instances for everything else, but there are some places where we haven't followed that convention.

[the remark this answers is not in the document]
alissapajer: The goal is just something readable for debugging and logging.

core/src/main/scala/quasar/datasource/url/reqTokenAuth/ReqTokenAuthMiddleware.scala
beckyconning: The original code is designed to get multiple values at the specified JSON path and this one is getting a single value from the specified JSON path. Does this requested change still apply given this?

alissapajer: You're right that the `take(1)` is a result of expecting a single value here, and thinking about it more, that component doesn't actually relate to what I was thinking about.

What I was thinking about is the scope in which we are going to maintain the resource, which in this case is the http connection. Using `client.run` uses `Resource` to contain that scope. `Resource` is a first class way to capture the allocation of a resource and effects to run upon resource finalization, and so in general it's our default "I need to capture a resource" data structure.

A singleton `Stream` is another data structure that can be used to capture a resource, but it is more of a second-class way to capture a resource. For one, `Resource` allows you to capture resource-related errors by forming a `MonadError`. Additionally, `Resource` allows you more control over a resource's scope and its finalizers.

@wemrysi is there any reason that flatmapping the request body into the outer singleton stream is going to cause any problems (like the current implementation is doing)? We're wedding the termination of the byte stream with the closure of the resource. If we ran `client.run`, we'd have to lift the result of compiling the inner stream into `Resource`, which would be essentially the same wedding. I can't convince myself I'm not missing anything though...

wemrysi: @alissapajer Since we know we want to do streaming things with the body, I think `client.stream` is appropriate here, providing convenience, resource safety and efficiency (we'll close the http connection as soon as possible). I agree that if we weren't going to consume the response body here, we'd consider returning the result in a `Resource`, but since we are consuming it we don't need to extend the resource scope outside of this function.

@beckyconning :golf:, `.compile.last.flatMap(_.fold(...` is a bit more direct

alissapajer: 👍 great, we can keep this using `client.stream`

core/src/main/scala/quasar/datasource/url/reqTokenAuth/ReqTokenAuthConfig.scala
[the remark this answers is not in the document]
beckyconning: awesome thank you!

core/src/main/scala/quasar/datasource/url/reqTokenAuth/ReqTokenAuthMiddleware.scala
beckyconning: awesome thanks!

README.md
```diff
@@ -44,7 +45,8 @@ Configuration for the URL datasource has the following JSON format
 * `throttling` (optional) the rate limit parameters
 * `meta` (optional) the metadata provided by frontend preserved without modifications
 * `variables` (optional) an object whose keys are variable names and values are `Variable`s
-* `oauth2` (optional) parameters to enable OAuth2 authentication for all requests
+* `oauth2` (optional) parameters to enable OAuth2 Authorization Code authentication for all requests
+* `reqTokenauth` (optional) parameters to enable OAuth2 Client Credentials authentication or similar non OAuth token request authentication for all requests
```
alissapajer, line 49: s/reqTokenauth/reqTokenAuth/

core/src/main/scala/quasar/datasource/url/oauth/OAuth2Middleware.scala
```diff
@@ -100,11 +46,31 @@ object OAuth2Middleware {
   private val ClientSecretParam = "client_secret"
   private val RefreshTokenParam = "refresh_token"
 
+  def apply[F[_]: MonadResourceErr: Sync](
+    store: ByteStore[F],
+    config: OAuth2Config)
+    (client: Client[F])
+    : Client[F] =
+      AuthMiddleware[F]("OAuth2", HttpAuthnStorageKey, httpAuth(config), store)(client)
```
alissapajer, line 54: new lined function input and return type should be four spaces indented, and the function body should be two spaces indented:

```scala
  def apply[F[_]: MonadResourceErr: Sync](
      store: ByteStore[F],
      config: OAuth2Config)
      (client: Client[F])
      : Client[F] =
    AuthMiddleware[F]("OAuth2", HttpAuthnStorageKey, httpAuth(config), store)(client)
````

File: core/src/main/scala/quasar/datasource/url/oauth/OAuth2Middleware.scala [hunk 10]
```diff
@@ -100,11 +46,31 @@ object OAuth2Middleware {
...
+
+  private def httpAuth[F[_]: MonadResourceErr: Sync](
+    config: OAuth2Config)
+    (store: ByteStore[F],
+    client: Client[F]) = {
```
alissapajer, line 59: indenting

core/src/main/scala/quasar/datasource/url/oauth/OAuth2Middleware.scala
```diff
@@ -100,11 +46,31 @@ object OAuth2Middleware {
...
+
   private def authorizationCode[F[_]: Sync](
-      store: ByteStore[F],
-      config: AuthorizationCodeConfig,
-      client: Client[F])
-      : F[Either[Throwable, HttpAuthentication]] = {
+    store: ByteStore[F],
+    config: AuthorizationCodeConfig,
+    client: Client[F])
+    : F[Either[Throwable, HttpAuthentication]] = {
```
alissapajer, line 73: indenting

core/src/main/scala/quasar/datasource/url/oauth/OAuth2Middleware.scala
alissapajer: huh, it says I marked this as resolved but I didn't and I can't see how to unmark it 🤷

core/src/test/scala/quasar/datasource/url/reqTokenAuth/ReqTokenAuthMiddlewareSpec.scala
```diff
@@ -0,0 +1,224 @@
...
+
+      underlying = Client.fromHttpApp(routes.orNotFound)
+      client = ReqTokenAuthMiddleware(bs, config)(underlying)
+
+      res <- MonadResourceErr[IO].attempt(client.fetch(GET(Uri.uri("http://example.com/baat")))(_.as[String]))
+      cachedToken <- bs.lookupString(HttpAuthnStorageKey)
+    } yield {
+      res must beLike {
+        case -\/(ResourceError.AccessDenied(_, Some(msg), _)) => {
+          msg must_=== "Parsing token request response as JSON failed."
```
alissapajer, line 217: I was surprised to see this error message leaking through. I'd expect the error message here to be `"Invalid credentials"` due to line 202 above. 

In `ReqTokenAuthMiddleware` we are doing `client.stream(req).flatMap(_.body)` to make the authentication request. This doesn't check if the request was a success or not, flatmapping on the body no matter what. This means that even if the request is not a success, we'll still go ahead and try to parse the body as json so we can gather the token. In this case, the body isn't valid json, so we error with this message. We should catch this error sooner by checking to see if the request was successful before parsing the body as json.

Will loop back with the best way to do that.

core/src/test/scala/quasar/datasource/url/reqTokenAuth/ReqTokenAuthMiddlewareSpec.scala
```diff
@@ -0,0 +1,224 @@
...
+      UrlForm.empty.updateFormField("grant_type", "client_credentials")
+
+    val routes = HttpRoutes.of[IO] {
+      case r @ GET -> Root / "baz" if isAuthd(accessToken, r) =>
+        Ok("data")
+
+      case r @ GET -> Root / "baz" =>
+        Unauthorized(`WWW-Authenticate`(Challenge("Bearer", "OAuth")))
+
+      case r @ POST -> Root / "oauth" / "token" =>
```
alissapajer, line 102: `AccessTokenSuccess` is only used in the implementation of OAuth2, so I was trying to figure out how this test was working.

This is how: we have `val ExamplePath = List(Object("access_token"))` set for these tests, which says that the token is expected to be found in json structured like this `{"access_token":"myaccesstoken"}`. The decoder for `AccessTokenSuccess` will decode its `AccessToken` into a field with name `"access_token"`.

We don't need to use `AccessTokenSuccess` to get the expected response here, and we also don't need the notion of `expectedParams`. 

Instead we can replace this case with
```scala
      case r @ POST -> Root / "oauth" / "token" =>
        Ok(s"""{"access_token":"$accessToken"}""")
```

Can you make this change in this test and the other tests which currently have `AccessTokenSuccess`.

Also can you add an additional test that has 
```scala
      case r @ POST -> Root / "oauth" / "token" =>
        Ok(s"""{"nope":"$accessToken"}""")
```
That will test what happens when we parse the json response, but there is no value at the config'd token path (which is `"access_token"`.

alissapajer: I made the change locally to make sure it would work, so decided to PR it to you https://github.com/precog/quasar-datasource-url/pull/515

Could you still add the additional test?

core/src/test/scala/quasar/datasource/url/reqTokenAuth/ReqTokenAuthMiddlewareSpec.scala
beckyconning: Request Token Auth is very broad especially compared to OAuth2 Code Authorisation. 

It doesn't understand or claim to understand the response other than that it was not JSON. However we should make more of the response that wasn't parsed as JSON shown in the error. I believe currently its quite truncated in the third argument of AccessDenied here.

beckyconning:
```
        case -\/(ResourceError.AccessDenied(_, Some(msg), Some(ParseException(err, _, _, _)))) => {
          msg must_=== "Parsing token request response as JSON failed."
          err must contain("Inva")
        }
```

This passes but with

```
          err must contain("Invalid credentials")
```

It doesn't because this is truncated by https://github.com/typelevel/jawn/blob/master/parser/src/main/scala/jawn/Parser.scala

beckyconning: Just realised what you mean here. Yep, seems we need two separate tests. One that should report that it was a bad request and another to report that it failed to parse as json. This will require a change to the reqtokenauthmiddleware. i can work this out. : )

core/src/test/scala/quasar/datasource/url/reqTokenAuth/ReqTokenAuthMiddlewareSpec.scala
```diff
@@ -132,7 +132,57 @@ object ReqTokenAuthMiddlewareSpec extends EffectfulQSpec[IO]
...
+
+      underlying = Client.fromHttpApp(routes.orNotFound)
+      client = ReqTokenAuthMiddleware(bs, config)(underlying)
+
+      res <- MonadResourceErr[IO].attempt(client.fetch(GET(Uri.uri("http://example.com/baz")))(_.as[String]))
+      cachedAuthn <- bs.getHttpAuthn(HttpAuthnStorageKey)
+    } yield {
+      res must beLike {
+        case -\/(ResourceError.AccessDenied(_, Some(msg), _)) => {
+          msg must_=== "Request token authentication failed: Couldn't find a value in the response body at the specified token path."
```
alissapajer, line 185: 👍

core/src/test/scala/quasar/datasource/url/reqTokenAuth/ReqTokenAuthMiddlewareSpec.scala
```diff
@@ -132,7 +132,57 @@ object ReqTokenAuthMiddlewareSpec extends EffectfulQSpec[IO]
...
+
+      underlying = Client.fromHttpApp(routes.orNotFound)
+      client = ReqTokenAuthMiddleware(bs, config)(underlying)
+
+      res <- MonadResourceErr[IO].attempt(client.fetch(GET(Uri.uri("http://example.com/baz")))(_.as[String]))
+      cachedAuthn <- bs.getHttpAuthn(HttpAuthnStorageKey)
+    } yield {
+      res must beLike {
+        case -\/(ResourceError.AccessDenied(_, Some(msg), _)) => {
+          msg must_=== """Request token authentication failed: Response body is not valid JSON. Body: {"access_token':"example"}"""
```
alissapajer, line 160: 👍 took me a minute to figure out what was invalid about the json :)

core/src/test/scala/quasar/datasource/url/reqTokenAuth/ReqTokenAuthMiddlewareSpec.scala
```diff
@@ -132,7 +132,57 @@ object ReqTokenAuthMiddlewareSpec extends EffectfulQSpec[IO]
     } yield {
       res must beLike {
         case -\/(ResourceError.AccessDenied(_, Some(msg), _)) => {
-          msg must_=== "Parsing token request response as JSON failed."
+          msg must_=== "Request token authentication failed: 400 Bad Request. Body: Invalid credentials"
```
alissapajer, line 135: 👍

core/src/main/scala/quasar/datasource/url/reqTokenAuth/ReqTokenAuthMiddleware.scala
```diff
@@ -68,20 +83,20 @@ object ReqTokenAuthMiddleware {
...
+            Some(s"Request token authentication failed: Response body was larger than the $InfoMaxSizeBytes byte limit."),
             None)))
 
       val noToken =
         MonadResourceErr[F].raiseError[HttpAuthentication](
-          ResourceError.connectionFailed(
+          ResourceError.accessDenied(
             ResourcePath.root(),
-            Some("Couldn't find any tokens at specified path"),
+            Some("Request token authentication failed: Couldn't find a value in the response body at the specified token path."),
```
alissapajer, line 95: 👍 to the more detailed error messages

core/src/main/scala/quasar/datasource/url/reqTokenAuth/ReqTokenAuthMiddleware.scala
```diff
@@ -0,0 +1,107 @@
...
+      }
+
+      def parseJson: Pipe[F, Byte, Json] = { bytes =>
+        bytes.chunks.parseJson[Json](AsyncParser.ValueStream) recoverWith {
+          case parseError: ParseException => {
+            bytes.through(utf8Decode).flatMap(body =>
+              Stream.eval(MonadResourceErr[F].raiseError[Json](
+                ResourceError.accessDenied(
+                  ResourcePath.root(),
+                  Some(s"Request token authentication failed: Response body is not valid JSON. Body: $body"),
```
alissapajer, line 75: Rendering the body here could leak sensitive credentials because the body may contain the invalid json (as part of the parse error) which contains the token. So while I agree with providing as much information as we can, we can't actually render the body here.

core/src/main/scala/quasar/datasource/url/reqTokenAuth/ReqTokenAuthMiddleware.scala
```diff
@@ -0,0 +1,107 @@
...
+    client: Client[F])
+    : F[HttpAuthentication] = {
+
+      val req: Request[F] =
+        UrlResource.request(config.req, config.headers)
+
+      def successBody(res: Response[F]): Stream[F, Byte] = {
+        if (res.status.isSuccess) { res.body } else {
+          val status = res.status.renderString
+          res.body.through(utf8Decode).flatMap(body =>
```
alissapajer, line 59: Do we expect the error messages from request token authentication to conform to `AccessTokenError.Standard` or would we always end up in `AccessTokenError.Unknown`? If we'd hit `Standard` then we could do `res.as[AccessTokenError]` here to get nicer formatting (and consistency with the oauth error messages).

core/src/main/scala/quasar/datasource/url/reqTokenAuth/ReqTokenAuthMiddleware.scala
beckyconning: I look forward to when we can begin a concerted effort towards improving "error" messaging. For example someone using the frontend to configure an API should be able to see the full HTTP requests and responses including bodies to aid in their configuration efforts and make such "errors" seem less like our fault and more like the next task the user needs to undertake to configure their datasource. In the meanwhile I'll remove the body from here.

core/src/main/scala/quasar/datasource/url/reqTokenAuth/ReqTokenAuthMiddleware.scala
beckyconning: I don't believe we this relates to AccessTokenError at all. Although Request Token Auth can be used to perform the OAuth2 Client Credentials flow it can also be used for far more things which don't fit into AccessTokenError or AccessTokenSuccess at all.

core/src/main/scala/quasar/datasource/url/reqTokenAuth/ReqTokenAuthMiddleware.scala
alissapajer: I agree with all you said.

I was thinking about this more, and in this case the token is visible in the data browser to anyone who has access to the datasource, so we should be able to show them the token in an error message. The problem is that the token is logged permanently, and while this _probably_ won't be a problem, we have to err on the side of caution when it comes to credentials.

core/src/test/scala/quasar/datasource/url/reqTokenAuth/ReqTokenAuthMiddlewareSpec.scala
```diff
@@ -143,7 +143,7 @@ object ReqTokenAuthMiddlewareSpec extends EffectfulQSpec[IO]
   "request fails with AccessDenied when access token not obtained due to non json response body" >>* {
     val routes = HttpRoutes.of[IO] {
       case r @ POST -> Root / "oauth" / "token" =>
-        Ok("""{"access_token':"example"}""")
+        Ok("""{"access_token":'<- single quote"}""")
```
alissapajer, line 146: that's awesome :) :)

core/src/main/scala/quasar/datasource/url/reqTokenAuth/ReqTokenAuthMiddleware.scala
alissapajer: Oh in this case we aren't actually logging the error message. But I still don't want to show it in the message as it's exposing more information to more people than were intended :(

PR #523, opened 2020-05-19
Initial requests should be discovered at the terminal request path

wemrysi:
During discovery, "initial" requests were being discovered at their initial URL path, which was unexpected from a user perspective. This changes the behavior so that they are now discovered at the path of the "terminal" part of the request.

This is a breaking change and could potentially cause tables previously based on an initial request to require reattaching to the terminal path if the "initial" and "terminal" paths differ. After discussion with @beckyconning and @garyb we determined the number of affected users is likely quite low and reattaching an affected table is easy to do.


wemrysi:
Released as `32.0.1-ddd7f3c`

alissapajer:
Not merging because 🛑 is up

wemrysi:
@alissapajer :stop_sign: is awaiting confirmation from @beckyconning that the issue prompting this PR is fixed with this change.

PR #531, opened 2020-05-26
Named datasets

beckyconning:


core/src/main/scala/quasar/datasource/url/ApiRequest.scala
```diff
@@ -123,6 +124,12 @@ object ApiRequest {
     }
   }
 
+  @SuppressWarnings(Array("org.wartremover.warts.Recursion"))
+  def getName(api: ApiRequest[_]): Option[String] = api match {
+    case Initial(a, b, c, term) => getName(term)
```
alissapajer, line 129: don't need recursion here; can do `term.name`

core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
```diff
@@ -124,15 +124,22 @@ class Http4sUrlDatasourceModule(clientBuilder: Http4sClientBuilder)
               case api :: List() =>
                 val path = ApiRequest.terminalUri(api)._2.singlePath(urlConfig.format)
                 val res = makeResult[F](client, api, urlConfig.headers, urlConfig.format)
-                IMap.singleton(path, res)
+                IMap.singleton(
+                  api.name.fold(path)(s => ResourcePath.root() / ResourceName(s)),
```
alissapajer, line 128: The path value here is used for error reporting like so, where the path is the `location`
```scala
        case Status.NotFound => // 404
          MonadResourceErr[F].raiseError[(Headers, EntityBody[F])](
            ResourceError.pathNotFound(location))
```
It's a disingenuous use of `PathNotFound` if we create the path according to the arbitrary string name given to the request. Ideally we could return a `PathNotFound` with the actual path _and_ the additional `api.name` information, when we have it. That way we we could give all the information we have to the frontend/user.

We need to look into how the frontend displays the `PathNotFound` message, since this would be a breaking change for the frontend.

Though, since this is urgent to land for today, I'm thinking that we can leave this be for now.

alissapajer:
pushing up some tests for the codecs...

alissapajer:
released as `33.0.4-c29f504`

core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
wemrysi: That's a good point, responding with a `ResourceError.ConnectionFailed` would probably be more appropriate than `PathNotFound` in any case. We could also get rid of the catch-all case in favor of `ConnectionFailed`.

core/src/main/scala/quasar/datasource/url/ApiRequest.scala
```diff
@@ -374,21 +383,34 @@ object ApiRequest {
...
+      } yield Basic(req, name)
+    }
 
   private def decodeBasic[U: DecodeJson]: DecodeJson[TerminalApiRequest[U]] =
-    DecodeJson(c => (c --\ "basic").as[UriRequest[U]].map(Basic(_)))
+    DecodeJson(c => {
+      val basic = (c --\ "basic")
+      for {
+        req <- basic.as[UriRequest[U]]
+        name <- (basic --\ "name").as[Option[String]]
```
alissapajer, line 402: We won't need to decode `"name"` in the legacy format because we only write in the current format.

alissapajer: Ok well I removed the `"name"` from the legacy decoders locally, and a test that claims to be decoding a current config format failed. There is no harm in keeping these here, so will do that. One more example of why we need versioned configs.

core/src/test/scala/quasar/datasource/url/ConfigSpec.scala
```diff
@@ -71,6 +71,40 @@ class ConfigSpec extends Specification {
...
+          |    "Accept": "*/*",
+          |    "Authorization": "admin:secret"
+          |  },
+          |  "requests": [
+          |    {
+          |      "basic": {
+          |        "url": "https://test.com",
+          |        "method": "POST",
+          |        "body": "test",
+          |        "name": "snaaaaaaake"
```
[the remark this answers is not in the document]
alissapajer: badger badger badger badger

core/src/main/scala/quasar/datasource/url/ApiRequest.scala
alissapajer: Actually, I just checked again, and it was actually a "backwards compat" check that failed. I had checked in a "backwards compat" test with a "name" decoding by accident.

alissapajer: @wemrysi opinion one way or the other on keeping these decoders? It's still not harmful, but we also don't need them.

alissapajer:
oh oops, should have been a **feature**, nbd

PR #541, opened 2020-06-05
implement reconfigure

alissapajer:
depends on https://github.com/precog/quasar/pull/4643


alissapajer:
@wemrysi I'm going to loop back to this before merge to write a lot more systematic tests, hitting all the various merge cases. I just threw a couple tests in there for now to convince myself that it's largely working.

alissapajer:
released as `34.0.6-88b1713`

core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
```diff
@@ -82,6 +82,54 @@ class Http4sUrlDatasourceModule(clientBuilder: Http4sClientBuilder)
...
+        patch.as[Config[String]].result match {
+          case Left(_) =>
+            Left(DatasourceError.MalformedConfiguration[Json](
+              kind,
+              sanitizeConfig(patch),
+              "Patch configuration to use in reconfigure is malformed."))
+          case Right(cfg) => Right(cfg)
+        }
+
+      patchIsSensitive =
```
wemrysi, line 107: What do you think about moving this and the actual merge implementation below to the `Config`? Then the semantics of "sensitive" would be defined there, which would help us keep them consistent when we make changes to the config itself.

alissapajer:
- [ ] TODO `"cookie"` and `"set-cookie"` are sensitive headers as well

core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
alissapajer: Sounds good to me!

alissapajer:
@wemrysi ready for another look!

core/src/main/scala/quasar/datasource/url/Config.scala
```diff
@@ -52,6 +56,35 @@ final case class Config[U](
...
+  def substituteNonURI(prefix: String): Config[U] = {
+    val unique = variables.foldLeft(self) { case (acc, subst) =>
+      subst.vary(prefix).substituteNonURI(acc)
+    }
+    variables.foldLeft(unique) { case (acc, subst) =>
+      subst.unvary(prefix).substituteNonURI(acc)
+    }
+  }
+
+  def reconfigureNonSensitive(patch: Config[U], kind: DatasourceType)
```
wemrysi, line 68: Eventually we might want to just return `Either[String, Config[U]]` (or maybe `Either[NonEmptyList[String], Config[U]]` if we want to list the specific sensitive inclusions). This would allow us to not have to include the `kind`, which is a bit awkward.

alissapajer:
released as `34.0.7-cb2631b`

PR #558, opened 2020-06-18
[ ] or { } value equals stop conditions include undefined and null 

alissapajer:
[ch11421]

broke up the commits for easier reviewing

commented out each line that got changed in the source to verify that a test fails in its absence

lumping empty structure, undefined, and null together satisfies the product requirements for now; at some point we may need to break them apart on the backend; see https://github.com/precog/slamx/pull/2770 for the new frontend text


alissapajer:
released as `35.0.0-ed8e9d1`

core/src/main/scala/quasar/datasource/url/pagination/ContinuationPlate.scala
```diff
@@ -36,6 +36,12 @@ import tectonic.Signal
  * If the type is `StopWhen`, `continue` will be set to false
  * if any value at the provided path matches the provided value,
  * even if that value is not seen last.
+ *
+ * If the `ContinuationValue` is `EmptyArray` (or `EmptyObject`),
+ * we treat `[]` (or `{}`), `null`, and undefined as the expected
+ * value. For example, if the value is `EmptyArray` and the condition
+ * is `ContinueWhile`, we continue while the value at the continuation
+ * path is `[]`, `null`, or absent (i.e. undefined).
```
djspiewak, line 44: This seems like a *really* strange semantic. Like, I think the code as-written correctly reflects it (obviously the tests back this up), but, it's a really really odd semantic. Is this really what we want?

alissapajer: Ultimately we may want to separate these three into three separate stop condition values, and then ask the user to choose which of them they want. This feature is this way for now because it was the fastest way to effectively add new stop conditions without having to change the codecs and change the frontend (neither of which are hard, but both of which take time).

@beckyconning as a second step, what do you think of separating this out into the these stop condition values:

- when `null`
- when `[]`
- when `{}`
- when undefined

alissapajer: As a side note, we'd need a UI design for the "when undefined" case. One idea would be a second button next to the "Is empty" button (which is what the user clicks to enter the `[]` and `{}` cases).

beckyconning: We need this for tomorrow incase we run into it during the marketo connector building.

I think a semantic that fits well into the data browser is "index is undefined".

E.g.
Data Page 1
```
{ "results": [{"name": "Alissa"},{"name":"Daniel"}] }
```
Data Page 2:
```
{ "results": [] }
```
Data browser:
| data    | results |
|-------------|---------------------------|
| All keys    | Indices:<br>Number |
| Key results | Values:<br>String  |
```

We pick "Indices: Number" and the stop condition value is "undefined"

If the value of the key `results` is `null` then the index is undefined, if the value of key `results` is `[]` then the index is undefined, if the key `results` is not present the index is undefined.

This is what we're trying to get at, there are no indices.

Doesn't matter why, but if there are no indices we're gonna stop loading rather than looping forever.

PR #564, opened 2020-07-04
add id to requests in config, make resources discoverable at their id path

alissapajer:
[ch11505]


alissapajer:
released as `35.2.2-04ddd19`

@garyb this PR adds the `id` field as an optional field, per request, meaning that the FE will not have to make any changes when this lands. This PR also exposes discovery+schema at the path `/precog_d645dc8f-e2dd-49ce-88c8-3e0108cef779/{request_id}`. When a datasource is saved with the new `id` values, each `id`'d request will be available for individual discovery+schema.

🛑 is up so we can continue to discuss how we're handling legacy formats of datasource configs, and make any changes to this PR if necessary.

alissapajer:
I'd like the make the `id` non-optional, but this will mean that the BE will have to generate new `UUID`s when it's encoding requests that don't have `id`s. The not ideal part of this is that this means that the results of a `GET /datasource/{id}` request will contain entries with the `id` field, but the discovery+schema endpoints won't work for that `id` until the FE persists the config. I spoke with @garyb about this, and this is no problem for the FE to do, as it will be doing this regularly when the user is editing a datasource. But, from the design perspective, the backend is serving sort of inaccurate information, plus we're directly conflating encoding configs with config pseudo-migration (pseudo because we're not actually persisting the updating config format).

Even if we do make the `id` optional, we're going to hit this same problem head-on when we pseudo-migrate the web datasource `initial` request into two separate requests, a parent and a child. The frontend will have to populate a column picker using the parent, but the `UUID` for this parent will not have been persisted in the actual on-disk configuration. Thus, the frontend will need to save the datasource (possibly as "invisible", see https://github.com/precog/features/pull/10/) before populating a picker.

One solution is a new `DatasourceModule#migrateConfig` endpoint. We do need to keep the scope of this limited though, so we may decide to delay this. But that would come at the cost of what we could call "API reasonability".

PR #566, opened 2020-07-07
Include id per request if missing

alissapajer:
part of ch11507

Requires the `id` field per-request in the internal representation of the datasource configuration (`Config`).

As a side-effect, all datasource configurations returns from `GET /datasource/{id}` and `GET /datasources` will include the `id` field, even if that `id` field is not present in the saved configuration. If the frontend wishes to use this `id` as the `path` in the discovery or schema endpoint, it will have to save the datasource with the `id` field present.

Since we still have to instantiate datasources using the configuration saved to disk (and we don't have config versioning and don't have a way to migrate plugin configurations on disk), it is still possible to persist a datasource without `id`s per request. But, whenever a datasource is saved it is best that `id` values be included.


alissapajer:
@rintcius @garyb I versioned this as **feature** because we're not actually able to enforce that datasources are saved with an `id` field per request, and thus the existing slamx code will still be compatible. Do you agree that this can be versioned as **feature** or is there a reason it should be **breaking**?

🛑 is up only in case this needs to be **breaking**.

garyb:
From talking on Discord, I'm fine with it being `feature`

core/src/main/scala/quasar/datasource/url/ApiRequest.scala
```diff
@@ -397,28 +397,25 @@ object ApiRequest {
   private def legacyDecodeLinkRelation[U: DecodeJson]: DecodeJson[TerminalApiRequest[U]] =
     DecodeJson( c => for {
       req <- c.as[UriRequest[U]]
-      id <- (c --\ "id").as[Option[UUID]]
       rel <- (c --\ "linkRelation").as[LinkHeaderRelation]
       name <- (c --\ "name").as[Option[String]]
-    } yield HeaderPaginated(id, req, rel, name))
+    } yield HeaderPaginated(UUID.randomUUID(), req, rel, name))
```
rintcius, line 402: `UUID.randomUUID()` needs to happen in `Sync[F]` since it isn't pure. Same in other places.

Maybe parameterize the function like this: `def legacyDecodeLinkRelation(id :UUID)` so that the decoders don't require the `F`. I think this also will pan out nicer in the tests, e.g. you don't have to replace id's anymore.

alissapajer: ah good catch thanks! will move that into an effect

alissapajer:
side-effecting json decoders have been giving me nightmares

so I'm closing this in favor of a proper (albeit versionless) datasource config migration

basic plan:
- add `migrateConfig` to `DatasourceModules`
- eagerly migrate the configs, per loaded plugin module, in the `Precog` or `Quasar` constructor

PR #578, opened 2020-07-22
dedup ResourcePath computation for error messages

alissapajer:


core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
```diff
@@ -581,34 +559,37 @@ class Http4sUrlDatasourceModule(clientBuilder: Http4sClientBuilder)
...
 
   private def makeResource[F[_]: ConcurrentEffect: MonadResourceErr](
       client: Client[F],
+      path: ResourcePath,
       req: UriRequest[Uri],
       headers: Map[CIS, String],
       format: DataFormat,
       name: Option[String])
       : Resource[F, (Headers, EntityBody[F])] = {
-    val path = name.fold(singleUrlPath(format, req.uri))(s => ResourcePath.root() / ResourceName(s))
```
alissapajer, line 596: This was the single call site for `singleUrlPath` (and for complete context, in `UrlResource.apply` the `path` is only used in the error messages). Instead of recomputing this path, we pass in the actual path used as the key in the `MapBasedDatasource`.

PR #579, opened 2020-07-22
Migrate requests in migrateConfig

alissapajer:


alissapajer:
This is technically **breaking** because a datasource can no longer be initialized with an out-of-date config. But, given that we migrate all datasources on application startup, we know that any `DatasourceRef` retrieved from storage will already be migrated. Additionally, we know that the FE is already encoding configs with the latest encoders, and so FE will not need to make any changes when this lands.

core/src/test/scala/quasar/datasource/url/ConfigSpec.scala
```diff
@@ -1283,6 +907,7 @@ class ConfigSpec extends Specification {
           42))
     }
 
+    // TODO move offset pagination migration into migrateConfig
```
alissapajer, line 910: Oh, there are these TODOs here.

alissapajer: Fixed.

alissapajer:
Thinking about what we want this to look like when we have actual versioning...

Versioned configs are one of the few places where I've found that true code duplicated (i.e. copy paste) is correct. There are a couple reasons for this

- the codecs for any particular version should _never_ change; if they need to change, a new version is added
- the precise codecs for each version should become transparent (no need to follow function calls or abstractions)
- it becomes trivial to delete no-longer-supported versions
- the overhead for adding a new version is very low: just copy-paste the previous version and then make your change

This does mean that there ends up being a lot of duplication in tests. Full test coverage is super important, so we should make sure it's easy to both add new tests relevant to the new version and run the existing tests against the new version when applicable. Need to think more about this aspect.

PR #584, opened 2020-07-29
move migration, reconfiguration, sanitization into their own spec files

alissapajer:


alissapajer:
force-pushed because I really want this change in a single commit for easier merging/rebasing against other local work

PR #586, opened 2020-07-30
Added custom authorization scheme

beckyconning:
Should this be feature or revision?


core/src/main/scala/quasar/datasource/url/oauth/OAuth2Middleware.scala
```diff
@@ -60,14 +61,15 @@ object OAuth2Middleware {
 
     config.grant match {
       case codeCfg: AuthorizationCodeConfig => {
-        EitherT(authorizationCode(store, codeCfg, client))
+        EitherT(authorizationCode(store, config.authorizationPrefix, codeCfg, client))
```
wemrysi, line 64: If we go with `BearerToken(scheme, token)`, then I think we'll want to default to "Bearer" here rather than passing the `Option` to `authorizationCode`.

core/src/main/scala/quasar/datasource/url/HttpAuthentication.scala
```diff
@@ -32,22 +33,31 @@ sealed trait HttpAuthentication extends Product with Serializable
 
 object HttpAuthentication {
   final case class BearerToken(token: String) extends HttpAuthentication
+  final case class OtherToken(name: CIS, token: String) extends HttpAuthentication
```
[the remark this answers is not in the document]
beckyconning: will this break all existing authorisations?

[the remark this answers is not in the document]
beckyconning: and does this matter?

beckyconning:
Potentially I need to do something here too https://github.com/precog/quasar-datasource-url/blob/master/core/src/main/scala/quasar/datasource/url/oauth/AccessTokenType.scala

beckyconning:
Zoho says it returns "Bearer" here https://www.zoho.com/crm/developer/docs/api/access-refresh.html I need to check if this is actually the case.

beckyconning:
```
{"access_token":"<REDACTED>","api_domain":"https://www.zohoapis.eu","token_type":"Bearer","expires_in":3600}
```

core/src/main/scala/quasar/datasource/url/oauth/OAuth2Middleware.scala
beckyconning: Should we do this in the codec or here?

core/src/main/scala/quasar/datasource/url/HttpAuthentication.scala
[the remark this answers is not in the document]
wemrysi: It shouldn't break any existing authorizations as the header sent in http requests wont change.

core/src/main/scala/quasar/datasource/url/oauth/OAuth2Middleware.scala
wemrysi: It will be both, in sense. In the codec, we'll add `"Bearer"` when decoding the existing tokens that don't specify a scheme. Here we're accounting for the fact that we don't require a scheme in the configuration and default to `"Bearer"` when it is omitted.

core/src/main/scala/quasar/datasource/url/HttpAuthentication.scala
```diff
@@ -24,30 +24,32 @@ import cats.syntax.eq._
...
 
-  def bearerToken(token: String): HttpAuthentication =
-    BearerToken(token)
+  def bearerToken(token: String, scheme: CIS): HttpAuthentication =
+    BearerToken(token, scheme)
 
   implicit val binaryCodec: Codec[HttpAuthentication] =
     discriminated[HttpAuthentication].by(uint8)
-      .| (0) { case BearerToken(t) => t } (BearerToken.apply) (utf8)
+      .| (0) { case BearerToken(t, s) => (t, s.value) } (tuple => BearerToken(tuple._1,CIS(tuple._2))) (utf8_32 ~ utf8)
```
wemrysi, line 48: Here we'll want to preserve the decoding of the previous `0` case, inserting `"Bearer"` for compatibility, adding a new case for the updated format

```suggestion
      .caseO[String](0)(_ => None)(t => BearerToken(t, CIS("Bearer"))) (utf8)
      .caseP(1) { case BearerToken(t, s) => (t, s.value) } (tuple => BearerToken(tuple._1, CIS(tuple._2))) (utf8_32 ~ utf8)
```

beckyconning: Smart!

PR #592, opened 2020-08-15
add Retry middleware to client

nicflores:
[ch11786]


alissapajer:
Looks like a legit failure in `Http4sUrlDatasourceModuleSpec`

core/src/test/scala/quasar/datasource/url/RetryMiddlewareSpec.scala
```diff
@@ -0,0 +1,111 @@
...
+import quasar.{NoopRateLimitUpdater, RateLimiter}
+import quasar.api.datasource.DatasourceError.InitializationError
+import quasar.api.resource.{ResourceName, ResourcePath}
+import quasar.connector.{ByteStore, MonadResourceErr, ResourceError}
+import quasar.connector.datasource.LightweightDatasourceModule
+import quasar.contrib.scalaz.MonadError_
+
+import scala.concurrent.ExecutionContext.Implicits.global
+
+object RetryMiddlewareSpec extends Specification
```
alissapajer, line 45: Could you add another test that shows that if the request doesn't succeed after the max retries is reached, it will return the last unsuccessful response?

core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
```diff
@@ -187,8 +188,29 @@ class Http4sUrlDatasourceModule(clientBuilder: Http4sClientBuilder)
             ZNel("Only one of 'oauth2', 'reqTokenAuth' or 'externalAuth' may be specified."))): MW).pure[F]
       }
 
+    val RetriableStatuses = Set(
+      //RequestTimeout,
+      // TODO Leaving PayloadTooLarge out until we model Retry-After
+      //InternalServerError,
+      //ServiceUnavailable,
+      //BadGateway,
+      Status.GatewayTimeout)
```
alissapajer, line 197: What's the reasoning for only supporting retries for `GatewayTimeout`?

core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
```diff
@@ -187,8 +188,29 @@ class Http4sUrlDatasourceModule(clientBuilder: Http4sClientBuilder)
...
+      }
+
     val configuredClient =
       clientBuilder[F]
+        .map(Retry(RetryPolicy(
+          RetryPolicy.exponentialBackoff(maxWait = 5.seconds, maxRetry = 5),
+          (req: Request[F], result: Either[Throwable, Response[F]]) => {
+            req.method.isIdempotent && isErrorOrRetriableStatus(result) 
+          }
+        )))
```
alissapajer, line 213: Can you move the parens to the previous line, so like 
```scala
          })))
```

core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
nicflores: I added the rest one by one to make sure nothing else would break.

alissapajer: Awesome. Now they they are all enabled, can you reference them from the http4s library, rather than duplicating them here?

Additionally you can use `RetryPolicy.defaultRetriable` to replace line 210 below.

https://github.com/http4s/http4s/blob/9b588daa6d371d960ba19f8eb6a76a5ae21ea3ec/client/src/main/scala/org/http4s/client/middleware/Retry.scala#L109

nicflores: Actually `InternalServerError` fails causes this test `raises an exception if other response is received` to fail.

core/src/test/scala/quasar/datasource/url/RetryMiddlewareSpec.scala
alissapajer: Another test request: could you add a test for each of the `RetriableStatuses`, showing that we retry with each them?

core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
```diff
@@ -187,8 +187,20 @@ class Http4sUrlDatasourceModule(clientBuilder: Http4sClientBuilder)
             ZNel("Only one of 'oauth2', 'reqTokenAuth' or 'externalAuth' may be specified."))): MW).pure[F]
       }
 
+    def isErrorOrRetriableStatus(result: Either[Throwable, Response[F]]): Boolean =
+      result match {
+        case Right(resp) => RetryPolicy.RetriableStatuses(resp.status)
+        case Left(_) => false
+        case _ => true
```
alissapajer, line 194: This case isn't reachable so it can be deleted.

Also, can you rename this to `isRetriableStatus` since it's not checking for errors anymore.

core/src/test/scala/quasar/datasource/url/Http4sUrlDatasourceModuleSpec.scala
```diff
@@ -1215,7 +1215,8 @@ object Http4sUrlDatasourceModuleSpec extends Specification {
 
       val dsmod = new Http4sUrlDatasourceModule(MockClientBuilder.applyErr(List(
         Right((Status.Ok, content, Headers.empty)),
-        Left(Left(new java.net.UnknownHostException("unknown"))))))
+        Left(Left(new java.net.UnknownHostException("unknown")))
+      )))
```
alissapajer, line 1219: dangling parens :) ):

PR #593, opened 2020-08-19
avoid validating config

nicflores:
[ch11800]


core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
```diff
@@ -195,19 +194,9 @@ class Http4sUrlDatasourceModule(clientBuilder: Http4sClientBuilder)
...
-          val ury = api.map(_._1)
-          UrlResource.request[F](ury.req, urlConfig.headers)
-        }
-
-      val ds = EitherT.fromEither[F](clientResult) flatMapF { client =>
-
-        val validated = validateFirstRequest(client, initialReqs) map {
-          case Condition.Abnormal(e) =>
-            Left(mapConfig(e)(κ(sanitizedJson)))
+      val ds = EitherT.fromEither[F](clientResult) flatMap { client =>
```
alissapajer, line 197: Now that we're not performing an effectful action while creating the datasource (previously the effectful action we were performing was making the initial http request), we're not benefiting from `EitherT`. `EitherT.fromEither` wraps the `Either` in `F.pure`, and there's no reason to wrap in `F.pure` unless you're in the context of `flatMap`ping over `F`, which we're not.

So, we can change this line to
```scala
val ds = clientResult map { client =>
```

Then line 207 becomes
```scala
new PrivateDatasource(precog_uuid, passInstructionThrough(ds))
```

and line 209 becomes
```scala
privateDatasource
```

Or at least something along those lines.

Then we can remove the value declarations `val ds` and `val privateDatasource` and just return their bodies directly, since that's what we'd be doing in any case.

core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
```diff
@@ -295,13 +283,13 @@ class Http4sUrlDatasourceModule(clientBuilder: Http4sClientBuilder)
...
-            Right(new PrivateDatasource(precog_uuid, passInstructionThrough(ds)))
+            new PrivateDatasource(precog_uuid, passInstructionThrough(ds))
         }
 
-        validated.recoverWith(datasourceInitErrorHandler(sanitizedJson).andThen(_.asLeft.pure[F]))
+        privateDatasource
       }
 
-      ds.value
+      ds.pure[F]
```
alissapajer, line 292: If you change `evalMap` to `map` on line 194, I think this can become just `ds`.

core/src/test/scala/quasar/datasource/url/Http4sUrlDatasourceModuleSpec.scala
```diff
@@ -1174,7 +1159,7 @@ object Http4sUrlDatasourceModuleSpec extends Specification {
       }
     }
 
-    "errors when GET to url is not found (mock-specific)" >> {
+    "Ok when GET to url is not found (mock-specific)" >> {
```
rintcius, line 1162: This test and some below are tests in the operation section, right? I think operation-time everything still need to behave as is, i.e. `beLeft` here and `ConnectionFailed` in the test below. 

The URL checks only needs to be avoided when the config is created or edited

core/src/test/scala/quasar/datasource/url/RetryMiddlewareSpec.scala
```diff
@@ -173,11 +173,9 @@ object RetryMiddlewareSpec extends Specification
           (Status.Ok, "42", Headers.empty))))
 
       useDatasource(dsmod, config)(_ => IO.unit)
-        .map(_ must beLike {
-          case Left(DatasourceError.ConnectionFailed(_, c, err)) =>
```
alissapajer, line 177: note to self so I don't forget - we still need this test to hit the retry limit

alissapajer:
@nicflores @rintcius pushed up a couple minor changes to the initialization and operation tests. This should address the concerns with the operation tests; does it all look correct?

@nicflores I also looked into the `RetryMiddlewareSpec` tests. Currently, all the tests in that file are using `useDatasource`. We need to switch them all to use `evaluateTyped`. The reason for this is that `useDatasource` does not actually evaluate the requests, and so in their current state, those tests aren't actually testing that we retry requests (since no requests are actually made). If we switch the tests to use `evaluateTyped`, then the tests will run actual requests, thus testing that the retry mechanic is retrying as expected.

Once the `RetryMiddlewareSpec` tests are changed to use `evaluateTyped`, I'm 👍 to merge.

core/src/test/scala/quasar/datasource/url/RetryMiddlewareSpec.scala
```diff
@@ -16,43 +16,58 @@
...
   type IE = InitializationError[Json]
 
+  def iRead[A](path: A): InterpretedRead[A] = InterpretedRead(path, ScalarStages.Id)
+
+  implicit final class DatasourceOps(val ds: LightweightDatasourceModule.DS[IO]) extends scala.AnyVal {
+    def evaluate(read: InterpretedRead[ResourcePath]) =
+      ds.loadFull(read) getOrElseF Resource.liftF(IO.raiseError(new RuntimeException("No batch loader!")))
+  }
+
   def useDatasource[A](
```
alissapajer, line 71: This isn't used here anymore, right?

core/src/test/scala/quasar/datasource/url/RetryMiddlewareSpec.scala
```diff
@@ -62,6 +77,22 @@ object RetryMiddlewareSpec extends Specification
         EitherT.fromEither[IO](r).semiflatMap(f).value
       })
 
+  def evaluateTyped(dsmod: Http4sUrlDatasourceModule, cfg: Json, name: String)
+      : IO[Either[IE, Either[String, Json]]] =
+    evaluateTyped(dsmod, cfg, ResourcePath.root() / ResourceName(name))
+
+  def evaluateTyped(dsmod: Http4sUrlDatasourceModule, cfg: Json, path: ResourcePath)
```
alissapajer, line 84: Do you need both of these `evaluateTyped` definitions?

alissapajer: Looks like you're only calling this one?

nicflores: Looks like I don't use that version. I'll clean it up.

core/src/test/scala/quasar/datasource/url/RetryMiddlewareSpec.scala
nicflores: I cleaned this one up in my last push.

core/src/test/scala/quasar/datasource/url/RetryMiddlewareSpec.scala
nicflores: Correct, I cleaned up the one above it.

rintcius (approved):
👍

PR #603, opened 2020-09-04
Add response body logging middleware

alissapajer:
[ch11817]


core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
```diff
@@ -189,6 +193,7 @@ class Http4sUrlDatasourceModule(clientBuilder: Http4sClientBuilder)
           })))
         .map(FollowRedirect(urlConfig.maxRedirects))
         .map(RateLimitingMiddleware(rl, _))
+        .map(LoggingMiddleware(10, s => ConcurrentEffect[F].delay(dsLogger.trace(s))))
```
alissapajer, line 196: enable with
```
      -
        name: quasar.datasource.url
        level: trace
```

alissapajer: We can make `10` a datasource config param if we need to later (kind of annoying that it'd have to be part of the datasource config, but is there another way?).

This feature is for internal use for debugging purposes, used by people writing integrations.

We're considering exposing a "debug" view in the product, which would contain request and response methods, headers, bodies. This would aid people writing integrations. But having this in the logs is fine for now.

alissapajer: Also to note: if you set the root level logger to TRACE
```
    Root:
      level: trace
```
all the bodies are logged. They're not printed in a particularly user-friendly format, but they are readable.

We're adding this middleware because the log files get too big if root level TRACE logging is turned on (since the logs will become the size of the response which could be GBs or more, and actually twice the size because they print both the byte value and the string char).

core/src/test/scala/quasar/datasource/url/LoggingMiddleware.scala
```diff
@@ -0,0 +1,84 @@
...
+import org.http4s._
+import org.http4s.client.Client
+import org.http4s.client.dsl.Http4sClientDsl
+import org.http4s.dsl._
+import org.http4s.implicits._
+
+import org.specs2.mutable.Specification
+
+object LoggingMiddlewareSpec extends Specification
+    with CatsIO
```
alissapajer, line 37: ok this is really really nice

nicflores (approved):

core/src/main/scala/quasar/datasource/url/LoggingMiddleware.scala
```diff
@@ -0,0 +1,68 @@
...
+      }
+
+    val back = s.pull.uncons flatMap {
+      case Some((hd, tl)) =>
+        Pull.eval(logChunk(hd)) >> Pull.output(hd) >> go(hd, tl)(1)
+      case None =>
+        Pull.done
+    }
+
+    back.stream
```
wemrysi, line 66: ⛳ using combinators
```scala
s.chunks.zipWithIndex flatMap {
  case (bytes, i) if i < max =>
    Stream.eval_(logChunk(bytes)) ++ Stream.chunk(bytes)
  case (bytes, _) =>
    Stream.chunk(bytes)
}
```

core/src/main/scala/quasar/datasource/url/LoggingMiddleware.scala
```diff
@@ -0,0 +1,68 @@
...
+import scala._, Predef._
+
+import cats.effect.{Resource, Sync}
+
+import fs2.{Chunk, Pull, Stream}
+
+import org.http4s.{Request, Response}
+import org.http4s.client.Client
+
+object LoggingMiddleware {
```
wemrysi, line 28: Maybe `ResponseBodyLoggingMiddleware`? nbd, but this is logging a very specific thing.

core/src/main/scala/quasar/datasource/url/Http4sUrlDatasourceModule.scala
wemrysi, line 196: Would we also want to include `org.http4s.client.middleware.Logger` to log the non-body parts of requests and responses?

Or if we only care about responses, there is `org.http4s.client.middleware.ResponseLogger`

alissapajer: Yeah @nicflores and I were looking at those as well. The headers and methods are being logged already by a netty logger, but if we log them with http4s we could enable/disable them via the logger with name `quasar.datasource.url`, which would be preferable.

core/src/main/scala/quasar/datasource/url/LoggingMiddleware.scala
alissapajer: @nicflores turns out `Pull` wasn't needed here since there are combinators (that themselves are implemented with `Pull` :) ) that do the job!

